PR #2, opened 2021-03-28
Add confirmation dialogs for file overwriting

vrmiguel:


src/dialogs.rs
```diff
@@ -0,0 +1,47 @@
...
+pub struct Confirmation<'a> {
+    pub prompt: &'a str,
+    pub placeholder: Option<&'a str>,
+}
+
+#[derive(Debug)]
+pub struct Error;
+
+impl<'a> Confirmation<'a> {
+    pub fn new(prompt: &'a str, pattern: Option<&'a str>) -> Self {
```
marcospb19, line 14: Could be useful for this structure to receive multiple `placeholder`s, so we could remove the `placeholder` Option, and instead insert multiple `"REPLACE"` in the `prompt`.

By this, when running `.ask()`, we accept a slice of patterns for replacing in order.

I don't know if we'll need this, so the current implementation looks fine.

However, I do think that "REPLACE" is better than providing a placeholder to simplify the code.

marcospb19 (approved):

Issue #13, opened 2021-04-05
Argparsing: problems with the `-o, --output` flag


vrmiguel:
[Relevant branch](https://github.com/vrmiguel/ouch/tree/misc)

* `ouch file.tar.gz -o new-folder`
    * Expected: create `new-folder` if it doesn't exist and decompress the files in there
    * What currently happens: `ouch` ignores the flag and saves to `pwd`

* `ouch file.tar.gz --output new-folder`
    * Expected: ouch src.tar.lzma --output new-src
    * What currently happens: ```[ERROR] cannot compress to 'new-src', likely because it has an unsupported (or missing) extension.```

marcospb19:
I'm debugging `-o` failing, is a problem with `oof`'s argparsing.
When using `--output` argparsing works, just like `-o` will when we fix it, so we need to debug that too.

Probably occurs inside of `Evaluator::compress_files`, when calling other functions.

marcospb19:
Apparently I forgot to fix that TODO.
![image](https://user-images.githubusercontent.com/38900226/113622180-fea9bc80-9632-11eb-9383-a7adc0bd7d61.png)

Now making the `-o` fix PR.

vrmiguel:
Solved by #14 and #16

PR #14, opened 2021-04-05
Fixing short flags not receiving values

marcospb19:
Fixes 50% of #13.


marcospb19:
(Yeah I expect to fail in here).

vrmiguel:
The issue is now solved on nightly

PR #22, opened 2021-04-09
Fix the -n, --no flag usage and add an alias for the compress subcommand

vrmiguel:


vrmiguel:
The build errors here seem to be related to internal GitHub Actions problems

PR #27, opened 2021-05-18
Add further testing to oof

fabricio7p:
.

### DIFF
--- a/oof/src/lib.rs
+++ b/oof/src/lib.rs
@@ -240,6 +240,19 @@ mod tests {
     }
 
     #[test]
+    fn test_unknown_short_flag() {
+        let flags_info = [
+            ArgFlag::long("output_file").short('o'),
+            Flag::long("verbose").short('v'),
+            Flag::long("help").short('h'),
+        ];
+
+        let args = gen_args("ouch a.zip -s b.tar.gz");
+        let result = filter_flags(args, &flags_info).unwrap_err();
+        assert!(matches!(result, OofError::UnknownShortFlag('s')));
+    }
+
+    #[test]
     fn test_pop_subcommand() {
         let subcommands = &["commit", "add", "push", "remote"];
         let mut args = gen_args("add a b c");
--- a/oof/src/lib.rs
+++ b/oof/src/lib.rs
@@ -234,11 +234,29 @@ mod tests {
     }
 
     #[test]
-    fn test_unknown_short_flag() {
+    fn test_unknown_flags() {
         let result = setup_args_scenario("ouch a.zip -s b.tar.gz c.tar").unwrap_err();
         assert!(matches!(result, OofError::UnknownShortFlag('s')));
         let result = setup_args_scenario("ouch a.zip --foobar b.tar.gz c.tar").unwrap_err();
-        // TODO: assert `UnknownLongFlag` error
+        // TODO: Try removing the `_` from `_foobar`
+        let _foobar = "foobar".to_string();
+        assert!(matches!(result, OofError::UnknownLongFlag(_foobar)));
+    }
+
+    #[test]
+    fn test_incomplete_flags() {
+        // TODO: Try removing the `_` from `_incomplete_flag`
+        let _incomplete_flag = Flag::long("output_file").short('o');
+        let result = setup_args_scenario("ouch a.zip b.tar.gz c.tar -o").unwrap_err();
+        assert!(matches!(result, OofError::MissingValueToFlag(_incomplete_flag)));
+    }
+
+    #[test]
+    fn test_duplicated_flags() {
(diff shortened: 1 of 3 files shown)


marcospb19:
Looking great, I will just request one change before merging.

oof/src/lib.rs
```diff
@@ -239,6 +239,19 @@ mod tests {
...
+    fn test_unknown_short_flag() {
+        let flags_info = [
+            ArgFlag::long("output_file").short('o'),
+            Flag::long("verbose").short('v'),
+            Flag::long("help").short('h'),
+        ];
+
+        let args = gen_args("ouch a.zip -s b.tar.gz");
+        let result = filter_flags(args, &flags_info).unwrap_err();
+        assert!(matches!(result, OofError::UnknownShortFlag('s')));
```
marcospb19, line 252: This procedure could be inserted inside of a closure or function (Rust supports functions inside of functions), so that you could test with other strings other `"ouch a.zip -s b.tar.gz"`, using the same `flags_info`, but with different args input that should return the same `OofError::UnknownShortFlag('s')`, that using multiple `assert!`s.

fabricio7p: Got it, will do

oof/src/lib.rs
```diff
@@ -220,6 +220,25 @@ mod tests {
...
 
+    fn setup_args_scenario(arg_str: &str) -> Result<(Vec<OsString>, Flags), OofError> {
+        let flags_info = [
+            ArgFlag::long("output_file").short('o'),
+            Flag::long("verbose").short('v'),
+            Flag::long("help").short('h'),
+        ];
+
+        let args = gen_args(arg_str);
+        filter_flags(args, &flags_info)
```
fabricio7p, line 231: compile error here, trying to return a value referencing local variable flags_info

fabricio7p (commented):
So i'm attempting to create this setup function you asked me. How would we tackle this compile error?
`filter_flags` needs a reference for `flags_info` and we cannot return them from within the function.
How would you solve this?

oof/src/lib.rs
marcospb19: Sorry for the late response, I'll take a look now.

fabricio7p: Ok please update this PR, i'm interested in seeing how to tackle this scenario

marcospb19: I made a commited into your branch, you can see the changes by clicking in `Files changed`, now that problem won't occur anymore.

oof/src/lib.rs
```diff
@@ -220,6 +222,43 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_unknown_flags() {
+        let result = setup_args_scenario("ouch a.zip -s b.tar.gz c.tar").unwrap_err();
+        assert!(matches!(result, OofError::UnknownShortFlag('s')));
+        let result = setup_args_scenario("ouch a.zip --foobar b.tar.gz c.tar").unwrap_err();
+        // TODO: Try removing the `_` from `_foobar`
```
fabricio7p, line 241: I don't quite understand what is happening here with this `matches!` macro. Please attempt to fullfill these TODOs and see what happens :smile:

fabricio7p (commented):
Added further testing. Getting weird compiler complains tho. I think it has to do with this matches macro

oof/src/lib.rs
marcospb19: `matches!` does not compare the fields, it only tries to pattern match against the enum variants, the argument you're passing, for example: `_duplicated_flag`, does not change the pattern.

Look at this this example:
```rust
let option: Option<i32> = Some(10);
match option {
    Some(value) => println!("{}", value),
    None => {},
}
```

`value` will be a binding to the value that is matched, now look at this other example:
```rust
let option: Option<i32> = Some(10);
let value = 102;
match option {
    Some(value) => println!("{}", value),
    None => {},
}
```

This code is exactly the same as the previous, `value` is not being used, the pattern is the same.

fabricio7p:
So i did a little refactor in the assert approach and added `test_misplaced_flag`.
Let me know if i'm doing good

marcospb19:
You're doing good! If you have finished let me know it is ready for merging.

For that we usually mark the PR as draft, and then unmark when it's ready, so I'll be marking your PR as draft here.

fabricio7p:
I think we're set. Every error variant is now covered.
This last one took me a while :laughing:

marcospb19:
It is possible that `oof` turns into a separated argparsing crate someday.

If you want to keep an eye on `ouch`, there's a button on the top right corner to receive notifications for some issues and PRs.

I created some other issues because it was a mess, if you need any mentoring again, don't be afraid to ask.

This PR is looking very good, thank you so much for the contribution.

PR #60, opened 2021-10-06
Don't use colors when `stdout` or `stderr` are being redirected

### LINKED ISSUE #59: Don't output colors if output is being redirect
By using [`atty`](https://docs.rs/atty) we can easily check if the `STDOUT` stream is pointing to another file descriptor.

Currently, redirecting `ouch` output to a file makes it unreadable in a lot of programs, as terminal escape sequences are only interpreted by terminals, and only by some tools (may work with `cat`, but not with `less` or `bat`).

Unresolved questions:
1. Should we add an `--color` flag to overwrite this behavior (and the NO_COLOR env var)?

Also, because we are already talking about colors.

2. We could also add `--no-color` flag to deactivate colors, maybe you are using a terminal that's not working properly with it?

vrmiguel:
Closes #59

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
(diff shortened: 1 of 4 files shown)


marcospb19:
Nice!

PR #84, opened 2021-10-08
Print the format type when the format is in an incorrent position

boozec:
When it raises an error caused by position of the format, now prints the format type as string instead of the position inside the array of formats.
In this way you can read on stdout the type like `.tar` or `.lz` instead of `1`, `2`, .., `n`.

## Example
Current code:
```
$ ./target/debug/ouch c hello.txt archive.tar.lz.tar.lz
[ERROR] Cannot compress to 'archive.tar.lz.tar.lz'.
 - Found the format '1' in an incorrect position.
 - 1 can only be used at the start of the file extension.

hint: If you wish to compress multiple files, start the extension with 1.
hint: Otherwise, remove 1 from 'archive.tar.lz.tar.lz'.
```

What fixs this PR:
```
$ ./target/debug/ouch c hello.txt archive.tar.lz.tar.lz
[ERROR] Cannot compress to 'archive.tar.lz.tar.lz'.
 - Found the format '.lz' in an incorrect position.
 - '.lz' can only be used at the start of the file extension.

hint: If you wish to compress multiple files, start the extension with '.lz'.
hint: Otherwise, remove '.lz' from 'archive.tar.lz.tar.lz'.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ For compiling, check [the wiki guide](https://github.com/ouch-org/ouch/wiki/Comp
 
 ## Supported formats
 
-|               | .tar | .zip | .bz, .bz2 | .gz | .xz, .lz, .lzma | .zst |
+|               | .tar, .tgz | .zip | .bz, .bz2 | .gz | .xz, .lz, .lzma | .zst |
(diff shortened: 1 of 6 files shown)


marcospb19:
Thanks for opening a PR!

PR #95, opened 2021-10-15
refactor: better NO_COLOR support

### LINKED ISSUE #62: The help message does not respect NO_COLOR

### LINKED ISSUE #66: Simplify colors implementation to remove code duplication
Each color could have it's function created with a macro, the function would receive the color &str as the argument and use conditional compilation inside in an `cfg!()` `if ` block, which is resolved and optimized out in compile time.

figsoda:
Fixes #62
Fixes #66

`colors::*` are now lazy statics that respect NO_COLOR instead of pure functions
- `colors::yellow()` -> `*colors::YELLOW`
- `if *NO_COLOR_IS_SET` is no longer needed outside of `colors` module

switched from lazy_static to once_cell since it doesn't use macros and the code can look a little cleaner

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -67,14 +67,14 @@ pub struct Bytes {
 pub mod colors {
     use once_cell::sync::Lazy;
 
-    static NO_COLOR_IS_SET: Lazy<bool> = Lazy::new(|| {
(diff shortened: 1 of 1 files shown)


figsoda:
Thanks for the review!
I edited to close that issue too

PR #99, opened 2021-10-15
Add proper error handling for conflicting flags

### LINKED ISSUE #55: Add proper error message when using conflicting flags (e.g. `--yes --no`)

SpyrosRoum:
This closes #55

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -11,7 +11,7 @@ use std::{
 
 use strsim::normalized_damerau_levenshtein;
 
-use crate::{arg_flag, flag, oof, Error};
+use crate::{arg_flag, error::FinalError, flag, oof, Error};
 
 #[derive(PartialEq, Eq, Debug)]
 pub enum Command {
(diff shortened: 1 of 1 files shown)


SpyrosRoum:
Honestly when you are making an app and not a library it doesn't often matter what the specific error is since most of the times you just want to return a nice message to the user and call it a day.
Personally when I am making an app I'd use something like [anyhow](https://crates.io/crates/anyhow) or [eyre](https://crates.io/crates/color-eyre).

In this particular case you can do what you said and if that's not enough then use one of the crates I mentioned earlier. You can still keep the builder pattern for the error but get rid of most of the other errors.

Btw if you need help with implementing this, or the move to clap you mentioned at some point then open an issue and let me know, I'd be happy to help.

marcospb19:
Thanks for the input!

@dlight suggested [`anyhow`](crates.io/anyhow) and [`eyre`](crates.io/eyre) too, I ended up being inspired by the usage of [`user-error`](crates.io/user-error) to create our own builder to display error with our custom syntax, error causes and the hints on how to fix it (the hints being inspired from `rustc` compile errors).

I'll make sure to open the issues :slightly_smiling_face:.

PR #101, opened 2021-10-17
Add support for short tar archive extensions

dnaka91:
Currently only `.tgz` as short form of `tar.gz` is supported but there are short versions for the other compression formats as well. This PR adds support for:

- `tar.bz` => `tbz`
- `tar.bz2` => `tbz2`
- `tar.xz` => `txz`
- `tar.lz` => `tlz`
- `tar.lzma` => `tlzma`
- `tar.zst` => `tzst`
﻿

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -246,10 +246,24 @@ fn compress_files(
                 writer.flush()?;
             }
             Tgz => {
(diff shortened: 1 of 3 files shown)


marcospb19:
Thanks for your contribution to `ouch`! We really appreciate it!

Issue #96, opened 2021-10-15
Panics when trying to compress folder with non-archive format


marcospb19:
Instead of panicking like this:

```sh 
# src/ is a folder
$ cargo run -q -- c src src.xz
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/commands.rs:70:85
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Should give a helpful error message, saying that its impossible to compress a folder with `.xz`.

Same applies for `.gz`, `.bz` and `.zst`.

marcospb19:
Fixed by #133.

Now:

```sh
$ cargo run -q -- c src src.xz
[ERROR] Cannot compress to 'src.xz'.
 - You are trying to compress multiple files.
 - The compression format 'xz' cannot receive multiple files.
 - The only supported formats that archive files into an archive are .tar and .zip.

hint: Try inserting '.tar' or '.zip' before 'xz'.
hint: From: src.xz
hint:  To : src..tarxz
```

PR #103, opened 2021-10-18
Remove unwrap for rfind

gwendalF:
Fixes https://github.com/ouch-org/ouch/issues/96

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -67,10 +67,18 @@ pub fn run(command: Command, flags: &oof::Flags) -> crate::Result<()> {
                 // Breaks if Lzma is .lz or .lzma and not .xz
                 // Or if Bzip is .bz2 and not .bz
                 let extensions_start_position = output_path.rfind(&extensions_text).ok_or_else(|| {
-                    Error::UnknownExtensionError(format!(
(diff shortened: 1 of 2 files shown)


marcospb19:
Thank you for contributing to `ouch`! We really appreciate it :slightly_smiling_face:.

Review with some requested changes coming soon.

src/commands.rs
```diff
@@ -66,7 +66,20 @@ pub fn run(command: Command, flags: &oof::Flags) -> crate::Result<()> {
 
                 // Breaks if Lzma is .lz or .lzma and not .xz
                 // Or if Bzip is .bz2 and not .bz
```
marcospb19, line 68: Hey, am back, I recently took a break from `ouch`.

So apparently #96 was fixed by changes made by @SpyrosRoum at #133, if you check this comments above the lines you edited, it references that might panic if extension is written in an alternate way, like LZMA or BZIP because those have these other aliases, `.lz` `.lzma` and `.xz` are the same thing for `ouch`.

marcospb19:
Because it's already fixed, I'll close this PR and update that comment by a safety one (like other unwraps in the code).

PR #118, opened 2021-10-20
change output flag to dir flag

### LINKED ISSUE #115: Changing `--output` to `--dir`
# Suggestion

Replace the `-o/--output` flag of the `decompress` subcommand by `-d/--dir`.

# Why?

```
OPTIONS:
    -o, --output <OUTPUT>    Decompress files in a directory other than the current
```

`--output` usually is referred to the file that is created, but in this case it's about the output directory.

Example: `gcc main.c --output main`.

Example: `ouch compress 1 2 3 archive.tar.gz`, 
If you check the _help_ for the subcommand `compress`, we the argument `archive.tar.gz` is the `<OUTPUT>` positional arg, again, has no relation to directories or folders, but output file instead.

khubo:
this changes the `-o/--output` to `-d/--dir` in decompress Command.

fixes #115

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,11 +34,11 @@ ouch decompress a.zip b.tar.gz c.tar
 ouch d a.zip
 ```
 
-You can redirect the decompression results to another folder with the `-o/--output` flag.
+You can redirect the decompression results to another folder with the `-d/--dir` flag.
 
 ```sh
 # Decompress 'summer_vacation.zip' inside of new folder 'pictures'
-ouch decompress summer_vacation.zip -o pictures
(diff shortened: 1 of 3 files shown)


marcospb19:
@GabrielSimonetto pointed out that this would not let us have `--output -` anymore if we want to use `STDOUT`. But we have other possible solutions for that problem.

In last instance, `/dev/stdout` and `/dev/stdin` with `--format` is guaranteed to solve it.

marcospb19:
Thanks for the contribution and sorry for the late merge, I took some time to rest recently.

PR #122, opened 2021-10-20
Generate completions and man page

figsoda:
Moved some parts of `cli.rs` to `opts.rs`, `ouch::cli::{Opts, QuestionPolicy, Subcommand}` are moved to `ouch::{Opts, QuestionPolicy, Subcommand}`

This uses a build script to generate completions when `GEN_COMPLETIONS=1` is present, this generates completions to `$OUT_DIR/completions`, for example:
- `GEN_COMPLETIONS=1 cargo build` would generate completions to `target/debug/build/ouch-*/out/completions`
- `GEN_COMPLETIONS=1 cargo build --release --target x86_64-unknown-linux-musl` would generate completions to `target/x86_64-unknown-linux-musl/release/build/ouch-*/out/completions`

A few more artifacts are generated in the completions in the `x86_64-musl` job:
- completions are generated and uploaded to `completions`
- man pages are generated with `help2man` and uploaded to `ouch.1`

Design choices:
- I chose completions to be generated to `OUT_DIR`, which is what rust/cargo recommends, but it also makes paths incredibly long and can possibly collide (when globbing with `*`) when completions are generated multiple times
- Man page could also be gzipped
- Completions could be in multiple artifacts, and powershell and zsh completions are named slightly intuitively since I left them as defaults

Side notes:
- I couldn't figure out a way for man pages to be generated inside `build.rs` since `help2man` excepts an executable, not a string
- The current ci is very big and repetitive, I have a very similar workflow that tests, builds, and creates releases on multiple platforms that we can probably borrow from: https://github.com/figsoda/mmtc/tree/main/.github/workflows

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -104,19 +104,26 @@ jobs:
       - name: Install dependencies for musl libc
         run: |
           sudo apt-get update
-          sudo apt-get install musl-tools
+          sudo apt-get install help2man musl-tools
 
       - name: Run cargo build
         uses: actions-rs/cargo@v1
         with:
           command: build
           args: --release --target x86_64-unknown-linux-musl
+        env:
+          GEN_COMPLETIONS: 1
 
       - name: Run cargo test
         uses: actions-rs/cargo@v1
         with:
           command: test
           args: --target x86_64-unknown-linux-musl
+
+      - name: Build man page and find completions
+        run: |
+          help2man target/x86_64-unknown-linux-musl/release/ouch > ouch.1
+          cp -r target/x86_64-unknown-linux-musl/release/build/ouch-*/out/completions .
 
       - name: Strip binary
         run: strip target/x86_64-unknown-linux-musl/release/ouch
@@ -127,6 +134,17 @@ jobs:
           name: 'ouch-x86_64-linux-musl'
           path: target/x86_64-unknown-linux-musl/release/ouch
 
+      - name: Upload completions
+        uses: actions/upload-artifact@v2
+        with:
+          name: completions
+          path: completions
+
+      - name: Upload man page
+        uses: actions/upload-artifact@v2
+        with:
+          name: ouch.1
+          path: ouch.1
 
   x86_64_glibc:
     name: Ubuntu 20.04 (glibc)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -115,6 +115,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_generate"
+version = "3.0.0-beta.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "097ab5db1c3417442270cd57c8dd39f6c3114d3ce09d595f9efddbb1fcfaa799"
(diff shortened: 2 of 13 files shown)


figsoda:
resolved the merge conflicts

marcospb19:
Nice PR!

> The current ci is very big and repetitive, I have a very similar workflow that tests, builds, and creates releases on multiple platforms that we can probably borrow from: https://github.com/figsoda/mmtc/tree/main/.github/workflows

Just saw your CI file and learned some stuff from it, seems like you can upload artifacts for each different job while keeping the different name for each of them, but can you optionally not upload the binaries for some of the jobs? (like we do currently).

> I chose completions to be generated to OUT_DIR, which is what rust/cargo recommends, but it also makes paths incredibly long and can possibly collide (when globbing with *) when completions are generated multiple times

The odds of this colliding is high?

> Man page could also be gzipped

Would be nice if you could gzip them CI before uploading, but I'm a noob when it comes to CI, is it easy to do? (@vrmiguel managed to learn github actions and did everything for `ouch`, I did practically nothing).

figsoda:
> seems like you can upload artifacts for each different job while keeping the different name for each of them

Actually I don't upload any artifacts, the jobs only upload stuff to the releases

> but can you optionally not upload the binaries for some of the jobs?

Should be possible, the main difference is more uses of matrix

> The odds of this colliding is high?

If you run `cargo build` with `GEN_COMPLETIONS` multiple times, trying to find the completion directory with `*`s will find you multiple directories, and there is no easy way to find out which is the newest one. This shouldn't be a issue in ci, but might be a little annoying when testing changes to completions

> Would be nice if you could gzip them CI before uploading

I should've mentioned that artifacts are in `.zip` formats, should I still gzip it into `ouch.1.gz` and have it be stored in a zip?

marcospb19:
> I should've mentioned that artifacts are in .zip formats, should I still gzip it into ouch.1.gz and have it be stored in a zip?

Oh yes, there's this, so no, leave them uncompressed, I'll probably need to create a script to help with releases anyway, so I could add this step to it.

figsoda:
resolved merge conflicts

marcospb19:
Resolved some more conflicts that appeared, thanks for the contribution and sorry for the delay!

marcospb19:
@figsoda have you tried out the zsh completion? Bash is working here but zsh is not.

figsoda:
I did not, I don't use zsh

Fish worked fine for me

yigitsever:
zsh completion is not working because of the following in `_ouch` (line 40 on my install);

```
':output -- The resulting file. It's extensions can be used to specify the compression formats:_files' \
```

The `'` in `It's` is causing the problem, replacing the line as;
```
":output -- The resulting file. It's extensions can be used to specify the compression formats:_files" \
```
Fixes the issue, I don't know about the generation process so just fixed it by hand in the meantime.

figsoda:
@yigitsever does this fix it for you?
https://github.com/ouch-org/ouch/pull/153

yigitsever:
@figsoda it does, didn't even realize the typo...

PR #125, opened 2021-10-21
Fix archive detection

### LINKED ISSUE #112: Wrong archive format detection patterns
# Bugs

Here's a bug: 

https://github.com/ouch-org/ouch/blob/2e6cd893dcde49fbb0cf7bb8719abf0069003edb/src/commands.rs#L58

Here's another:

https://github.com/ouch-org/ouch/blob/2e6cd893dcde49fbb0cf7bb8719abf0069003edb/src/commands.rs#L195

# Reason

In both cases we are pattern matching to see if the file is an archive or not.

Formats that are archives: `tgz`, `tar`, `zip`, `tzst`, `zstd`, etc.
No archives: `xz`, `gz`, `bz`, `zst`.

However, we added new variants and forgot to update these patterns to match.

https://github.com/ouch-org/ouch/blob/2e6cd893dcde49fbb0cf7bb8719abf0069003edb/src/extension.rs#L8-L20

# Solution

Add method to `CompressionFormat` called `is_archive_format` and use it and it's negation instead.

Then check for each non-exhaustive pattern matching done in `ouch` with `if let`, `matches!` and `match`. Reviewing if should be replaced by a call of `format.is_archive_format()` (or `!format.is_archive_format()`)

SpyrosRoum:
This introduces an `is_archive_format` method on `CompressionFormat` and uses it were appropriate 

This closes #112

### DIFF
--- a/src/extension.rs
+++ b/src/extension.rs
@@ -21,7 +21,7 @@ pub enum CompressionFormat {
 
 impl CompressionFormat {
     pub fn is_archive_format(&self) -> bool {
-        matches!(self, Tar | Tgz | Tbz | Tlzma | Tzst | Zstd | Zip)
+        matches!(self, Tar | Tgz | Tbz | Tlzma | Tzst | Zip)
     }
 }
 
--- a/src/extension.rs
+++ b/src/extension.rs
@@ -19,6 +19,12 @@ pub enum CompressionFormat {
     Zip,   // .zip
 }
 
+impl CompressionFormat {
+    pub fn is_archive_format(&self) -> bool {
+        matches!(self, Tar | Tgz | Tbz | Tlzma | Tzst | Zstd | Zip)
+    }
+}
+
 impl fmt::Display for CompressionFormat {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(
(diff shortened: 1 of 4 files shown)


SpyrosRoum:
Okay, I believe up to this everything is good, tests will be complaining though because it's trying to compress to `tar.zst` but `zst` is an archive format and we don't allow that kind of thing currently. Is this correct, can I just remove these cases from the tests?

marcospb19:
Nice PR, sorry for the late merge, I took some time from `ouch` but will try to come back now.

src/extension.rs
```diff
@@ -19,6 +19,12 @@ pub enum CompressionFormat {
     Zip,   // .zip
 }
 
+impl CompressionFormat {
+    pub fn is_archive_format(&self) -> bool {
+        matches!(self, Tar | Tgz | Tbz | Tlzma | Tzst | Zip)
```
marcospb19, line 24: I wonder if this non-exhaustive pattern matching (with the implicit `_` wildcard) is a great choice, because we can forget to add a new format to this list.

Did not questioned this in the review process because I know people consider not matching every variant very convenient.

SpyrosRoum: I agree, proper match without any `_` would be better in the long term.
Though since this has been merged I urge you to take a look at #133

PR #130, opened 2021-10-29
Print number of unpacked files

dcariotti:
When you use the decompression tool you can avoid to overwrite one (or more) files. `crate::archive::tar::unpack_archive` already returns a list of the unpacked files, so after the "Success" information,﻿ it also prints the number of successfully unpacked files.


marcospb19:
ty!

Issue #88, opened 2021-10-14
Compression error hints can make wrong suggestions


marcospb19:
_Originally posted by @dcariotti in https://github.com/ouch-org/ouch/pull/84#discussion_r726548754_

# Problem

After running this:
```sh
ouch c 1.txt 2.txt archive.lz.tar
```

Ouch shows an error message:
```c
[ERROR] Cannot compress to 'archive.lz.tar'.
 - You are trying to compress multiple files.
 - The compression format '.lz' cannot receive multiple files.
 - The only supported formats that archive files into an archive are .tar and .zip.

hint: Try inserting '.tar' or '.zip' before '.lz'.
hint: From: archive.lz.tar
hint:  To : archive.tar.lz.tar
```

However, if you follow what the hints suggest and use `archive.tar.lz.tar`:

```sh
ouch c 1.txt 2.txt archive.tar.lz.tar
```

You will receive the following error:
```c
 - Found the format '.tar' in an incorrect position.
 - '.tar' can only be used at the start of the file extension.

hint: If you wish to compress multiple files, start the extension with '.tar'.
hint: Otherwise, remove '.tar' from 'archive.tar.lz.tar'.
```

---

Our hints are not ready for this specific errors.

Maybe it probably should just have said "you might want to move `.tar` to the start".

There are actually a lot of ways we can rethink those error messages to fix this issue, I'm not currently sure what's the best one.

PR #133, opened 2021-10-31
Remove tar combinations from compression format

SpyrosRoum:
This allows us to remove a lot of duplicate code and make things simpler. It also doesn't remove any functionality as far as I can see.


marcospb19:
It's been some time since I have been thinking about a rewrite of `CompressionFormat`, I planned on something like this:

```rust
pub struct Extension {
    compression_formats: Vec<CompressionFormat>,
    display_text: String, // or &str or OsStr or OsString, idk
}
```

So the extension `.tgz` would be made of:

```rust
Extension { compression_formats: vec![Tar, Gz], display_text: "tgz".into() }
```

Cause the code in this PR will likely break this piece of code:

https://github.com/ouch-org/ouch/blob/a02eb452c0732f6eaa397fa95072c1506568f101/src/commands.rs#L63

In this line, we get all the compression formats and use them to compose this string of extensions, so we can have the flexibility to change `archive.gz.xz` to `archive.tar.gz.xz` by just adding that `.tar` in the middle.

This requires the `Display` impl:
https://github.com/ouch-org/ouch/blob/a02eb452c0732f6eaa397fa95072c1506568f101/src/extension.rs#L28-L47

Of `Extension` to retrieve what was given in the CLI.

For example: If the user typed `archive.tgz`, and we trigger some error message with this extensions Display stuff, it would print `archive.tar.gz` instead, which can be confusing. 

# Solutions:

1. Rework the `Extension` and `CompressionFormat` as I told in this comment.
2. Change how we use extensions in error messages, having another system of "parsing the file_name of a filepath".
3. Other?? idk

(See https://github.com/ouch-org/ouch/issues/88 too)

SpyrosRoum:
Is this about what you were thinking? 
It touches a lot of sensitive places so I'd appreciate it if you could take a good look (we should really improve our tests soon).

figsoda:
or we could just not store the extension strings and get it from the original path every time we want to print an error

marcospb19:
> or we could just not store the extension strings and get it from the original path every time we want to print an error

That's also possible, but in that case we would need another system to do the parsing and reformatting again, could happen, but may be harder to implement.

marcospb19:
Merged with some warnings still, but I can fix those.

Maybe the missing_docs lints is too aggressive, idk.

marcospb19:
Oops, found one bug:

```
$ cargo run -q -- c src src.xz
[ERROR] Cannot compress to 'src.xz'.
 - You are trying to compress multiple files.
 - The compression format 'xz' cannot receive multiple files.
 - The only supported formats that archive files into an archive are .tar and .zip.

hint: Try inserting '.tar' or '.zip' before 'xz'.
hint: From: src.xz
hint:  To : src..tarxz
```

The last line:

```
hint:  To : src..tarxz
```

Should be `src.tar.xz`.

SpyrosRoum:
Well that case used to panic before so overall it's an improvement I'd say haha

marcospb19:
Bug 2 comments above was fixed by #219 :slightly_smiling_face:.

PR #135, opened 2021-11-01
rewrite ci

figsoda:
- builds and tests in debug mode with stable
- builds in release mode and uploads stripped binaries with nightly
- uploads completions and man page on x86_64-unknown-linux-musl
- checks with clippy and rustfmt

platforms:
- x86_64-apple-darwin
- x86_64-pc-windows-msvc
- x86_64-unknown-linux-gnu
- x86_64-unknown-linux-musl


marcospb19:
> builds in release mode and uploads stripped binaries with nightly

Maybe we should upload binaries compiled with stable Rust while we don't use any proper Rust nightly feature, just the `rustfmt` unstable features.

This way CI can also detect if we are using nightly features.

Not that I want to block nightly, it has nice features but we would only like to depend on it if given a big reason.

figsoda:
nightly rust supports stripping so we don't have to figure out how to strip binaries on different platforms

marcospb19:
Ok, that's very useful :smile: merging!

PR #138, opened 2021-11-02
Simplify cli canonicalize implementation

marcospb19:
The FileNotFound error should be treated in the `crate::Error::From<io::Error>` conversion implementation.


marcospb19:
Opened by mistake with a lot of other stuff.

PR #141, opened 2021-11-02
Fix decompression overwritting files without asking and failing on directories

### LINKED ISSUE #136: Decompression error when folder already exists at the place of output file
image

Ouch doens't really know that there is a folder there, it just gives this error message.

When it's not a folder, but a simple text file, it does ask if should overwrite.

SpyrosRoum:
Closes #136

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -326,22 +326,12 @@ fn decompress_file(
         Gzip | Bzip | Lzma | Zstd => {
             reader = chain_reader_decoder(&formats[0].compression_formats[0], reader)?;
 
-            let mut writer = match fs::OpenOptions::new().write(true).create_new(true).open(&output_path) {
-                Ok(w) => w,
-                Err(e) if e.kind() == io::ErrorKind::AlreadyExists => {
-                    if utils::user_wants_to_overwrite(&output_path, question_policy)? {
-                        if output_path.is_dir() {
-                            // We can't just use `fs::File::create(&output_path)` because it would return io::ErrorKind::IsADirectory
-                            // ToDo: Maybe we should emphasise that `output_path` is a directory and everything inside it will be gone?
-                            fs::remove_dir_all(&output_path)?;
-                        }
-                        fs::File::create(&output_path)?
-                    } else {
-                        return Ok(());
-                    }
-                }
-                Err(e) => return Err(Error::from(e)),
-            };
+            let writer = utils::create_or_ask_overwrite(&output_path, question_policy)?;
+            if writer.is_none() {
+                // Means that the user doesn't want to overwrite
+                return Ok(());
+            }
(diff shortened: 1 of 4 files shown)


src/commands.rs
```diff
@@ -326,8 +326,22 @@ fn decompress_file(
...
+                            // ToDo: Maybe we should emphasise that `output_path` is a directory and everything inside it will be gone?
+                            fs::remove_dir_all(&output_path)?;
+                        }
+                        fs::File::create(&output_path)?
+                    } else {
+                        return Ok(());
+                    }
+                }
+                Err(e) => return Err(Error::from(e)),
+            };
```
marcospb19, line 344: Maybe if you move this to utils.rs, you can reuse the same code for the .zip and .tar parts?

Would that be possible? something like `create_file_checking_overwrites()` which receives one AsRef<Path> and tries to open with those OpenOptions configs you used, and returns `Result<Option<File>>`, with `Ok(None)` in case the used does not want to overwrite.

marcospb19 (changes requested):

src/commands.rs
SpyrosRoum: Yeah that's a good point

SpyrosRoum: I don't think this is really doable after all because the archives don't want the file to exist there at all. Good news is I caught a bug because of this.
I did extract the function anyway because I think it's a bit cleaner but I didn't end up using it in .zip and .tar
It *could* be used if I were to add some flags determining if the file should be created or not but I don't think it's worth it, what do you think?

marcospb19: I don't think that's worth too, but by how that function looked, I think it's better now on utils.rs :D .

marcospb19 (approved):

marcospb19:
Awesome that you got another bug in this, makes me wonder how many specific errors like this still exist in `ouch`...

SpyrosRoum:
With this out of the way I am thinking it might be time for a new release?
We have fixed many bugs, some of them pretty critical..

marcospb19:
> With this out of the way I am thinking it might be time for a new release?

Yeah, it's been a long time since I said I was going to make a new release, we had too many changes.

I'm not sure what to put in the changelog, I think I'll make it more organized after the 1.0 release.

SpyrosRoum:
> I'm not sure what to put in the changelog, I think I'll make it more organized after the 1.0 release.

I haven't used it personally but take a look at [realease-please](https://github.com/googleapis/release-please), it looks perfect for that kind of thing

marcospb19:
I think it will get easier when ouch leaves the unstable stage, let's see, I'm not a huge fan of pushing extra requirements like the "conventionalcommits" because I prefer to keep it simple for people to contribute.

PR #150, opened 2021-11-02
Add support for lz4

### LINKED ISSUE #120: LZ4 support?
Any plans to support LZ4 as well? There are some high-quality crates that could provide the functionality.

The algorithm itself is one of the fastest when it comes to decompression time. I came across a [benchmark](https://github.com/inikep/lzbench#benchmarks) that showed it was only about half as slow as memcpy in terms of data throughput. It sacrifices compression ratio to meet those throughput requirements.

In another [benchmark](https://github.com/lz4/lz4#benchmarks) in liblz4's README, it compresses the Silesia Corpus faster than anything else and decompresses it faster than anything else as well. Zlib can achieve a higher compression ratio but will take 21x longer to get the size down 1.47x.

LZ4's inventor, Yann Collet, also build ZStandard and continues work on both of them to this day.

It's a popular compression scheme in the Hadoop world. I'm sure as your CLI's userbase grows it'll become more popular as well. It would be great to offer people the choice between faster compression and decompression times or compression ratio.

figsoda:
Closes #120

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -256,6 +256,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7fb9b38af92608140b86b693604b9ffcc5824240a484d1ecd4795bacb2fe88f3"
 
 [[package]]
+name = "lz4_flex"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


figsoda:
I changed `_` to `Tar | Zip` so It would be easier to add new formats looking at the error messages

PR #153, opened 2021-11-02
fix typo

figsoda:
I think this also fixes the zsh completion issue because clap didn't escape the single quote @marcospb19 
probably related: https://github.com/clap-rs/clap/issues/1596


marcospb19:
Will probably do a new release with this one.

figsoda:
maybe, yeah

Issue #145, opened 2021-11-02
Try to infer file extension as fallback when decompressing


sigmaSd:
Here is a code example: https://github.com/sigmaSd/ouch/commit/282324455bded82ac007e3fe66736e90de77a857

marcospb19:
Related to #142

sigmaSd:
The usecase Im thinking is a file without extention

Ouch can print something like: "Detected file type tar , do you want to decompres it?"

But I dont know how common this use case is (maybe not at all)

Feel free to close this issue if you think this should not be supported

marcospb19:
It's really similar to what I proposed in #142, I think it can be a great addition, but I'd ask you to first implement it as a warning using the `info!` macro.

This warning would be emitted even if the file has extensions, but the implementation would be very similar.

About if the [Y/n] question should or not be asked in the **specific case** when no extensions are given, I'd like to wait on @vrmiguel and others opinion.

sigmaSd:
Also another idea is when Ouch is done decompressing, it should detect the detect if the decompressed file is also an archive and prompt the user to decompress it if so.

marcospb19:
You need to add "fixes #145" in the PR to link them and automatically close issues on merge.

https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

sigmaSd:
I see thanks I'll keep that in mind.

Issue #142, opened 2021-11-02
Check MIME types before decompressing to warn mismatching formats


marcospb19:
When decompressing, we usually rely on the file extensions being correct ("usually" because #134 will be added).

We should warn the users when the MIME type of some time does not match the decompression method we are using, giving an `[Y/n] dialog to ask if they want to proceed or skip.

Because of implementation limitations and complexity, we should check the MIME only at the surface of the file, even tho `ouch` supports files like `archive.tar.gz.xz.bz2.zst.lz`, our warnings should not decompress all the steps for checking MIME, only what can be read directly.

Implementation details: implementation should save the bytes in a small Vec/array instead of reading from the same file twice, because we also plan on support reading from STDIN, and you cannot read from it twice.

(Previously #5.)

PR #154, opened 2021-11-03
Infer file extension when decompressing

sigmaSd:
https://github.com/ouch-org/ouch/issues/142
https://github.com/ouch-org/ouch/issues/145


marcospb19:
Thanks for your first contribution to `ouch`! We really appreciate it, review will be coming the next days (probably tomorrow).

src/commands.rs
```diff
@@ -473,3 +485,38 @@ fn list_archive_contents(
...
+            let outer_ext = format.iter().next().unwrap();
+            if outer_ext != &detected_format {
+                info!("The file extension: `{}` differ from the detected extension: `{}`", outer_ext, detected_format);
+                if !user_wants_to_continue_decompressing(path, question_policy)? {
+                    return Ok(ControlFlow::Break(()));
+                }
+            }
+        } else {
+            // Could not detect the extension automatically
+            // Should we tell the user about it?
```
marcospb19, line 518: I guess you don't need to tell the user about it because it's guaranteed that `!format.is_empty()`, right?

src/utils/fs.rs
```diff
@@ -80,6 +80,77 @@ pub fn nice_directory_display(os_str: impl AsRef<OsStr>) -> Cow<'static, str> {
...
+        buf.len() > 3 && buf[0] == 0x04 && buf[1] == 0x22 && buf[2] == 0x4D && buf[3] == 0x18
+    }
+    fn is_zst(buf: &[u8]) -> bool {
+        buf.len() > 3 && buf[0] == 0x28 && buf[1] == 0xB5 && buf[2] == 0x2F && buf[3] == 0xFD
+    }
+
+    let buf = {
+        let mut b = [0; 270];
+        // Reading errors will just make the inferring fail
+        let _rn = std::fs::File::open(&path).map(|mut f| std::io::Read::read(&mut f, &mut b));
```
marcospb19, line 128: What `_rn` means?

src/utils/fs.rs
```diff
@@ -80,6 +80,77 @@ pub fn nice_directory_display(os_str: impl AsRef<OsStr>) -> Cow<'static, str> {
...
+            && (buf[3] == 0x4 || buf[3] == 0x6 || buf[3] == 0x8)
+    }
+    fn is_tar(buf: &[u8]) -> bool {
+        buf.len() > 261
+            && buf[257] == 0x75
+            && buf[258] == 0x73
+            && buf[259] == 0x74
+            && buf[260] == 0x61
+            && buf[261] == 0x72
+    }
```
marcospb19, line 99: I wonder, what are the differences between doing this and using a crate like `infer`?

src/utils/fs.rs
```diff
@@ -80,6 +80,77 @@ pub fn nice_directory_display(os_str: impl AsRef<OsStr>) -> Cow<'static, str> {
     }
 }
 
+/// Try to detect the file extension by looking for known magic strings
+pub fn try_infer(path: &Path) -> Option<Extension> {
+    fn is_zip(buf: &[u8]) -> bool {
+        buf.len() > 3
+            && buf[0] == 0x50
```
marcospb19, line 87: Where you got the information about these magic numbers from? Could you add a comment in code so that we can refer to it later? (or send it here and I'll make sure to add it later).

src/utils/fs.rs
sigmaSd: It is the same as infer (because I copied code from it), except infer is missing is_lz4 and is_zst

src/utils/fs.rs
sigmaSd: https://en.wikipedia.org/wiki/List_of_file_signatures

src/utils/fs.rs
sigmaSd: I can just change it to _ ( I think in my mind was something like result number)

src/utils/fs.rs
sigmaSd: Infer: https://github.com/bojand/infer/blob/3221954081587a8dbaee83f0b98a1bee49c54193/src/lib.rs#L236

src/commands.rs
sigmaSd: That else branch is this case:
- The file has an extentiom
- We tried to detect the extention but it failed

The most probable case is that the file have an extention that we dont support

I can just add a simple warning

sigmaSd: What I want to write is something like:
`warning!("Could not detect the file extension, are you sure that this is a valid archive?");`

But the only problem is I'm not sure if there are some false positive or not.

sigmaSd:
I updated the pr, I can squash the commits later when the review is done.

src/commands.rs
marcospb19: I also fear we run into some false positives, so I think in this case we should just warn the user using the `info!()` macro, and proceed normally, if our warning is wrong, at least everything still works.

src/utils/fs.rs
marcospb19: Nice, I'll be copying the LICENSE copyright notice.

src/commands.rs
sigmaSd: done

marcospb19:
Sorry for my late review.

> I updated the pr, I can squash the commits later when the review is done.

I left one conversation unresolved for that warning, before you squash it.

sigmaSd:
I added the info!

marcospb19:
I think I can squash that for you.

marcospb19:
I found a small bug at the zip detection and fixed it, unfortunately git history says I commited and you authored, I don't understand why it's not the opposite, but anyways, thanks for your contribution! This is a great addition.

sigmaSd:
Thanks as well!

That function is copied from infer, so I'll let the author know.

marcospb19:
For reference, you sent the https://en.wikipedia.org/wiki/List_of_file_signatures link, there, we have the line:

50 4B 03 04
50 4B 05 06 (empty archive)
50 4B 07 08 (spanned archive)

marcospb19:
(There is also a chance that our source contains the wrong information and the original code was right all along. Pinging @bojand here too)

PR #163, opened 2021-11-04
rewrite tests

### LINKED ISSUE #30: Add nested folders to automated compression and decompression tests.
Currently, we only test archive formats with multiple regular random files, there's no nested folders testing, so for that formats the tests need to be refactored.

### LINKED ISSUE #31: Test compression and decompression of single file compression formats
Some formats, like .xz can only compress one single file, as opposed to .zip that compresses multiple and bundle them together in one file.

Currently we are only testing formats that create archives, and throwing single file on top of it.

Maybe we should test the single file formats separately.

`.tar.xz` already indirectly tests for `.xz`, that is enough for compression and decompression purposes, but we want to make sure that the CLI won't panic in any specific case.

### LINKED ISSUE #50: Having multiple archive tests running at once results in crazy crashes
This issue **DOES NOT** affect any release of `ouch`, just running the development tests, however, it can turn out to be a pain for us developers in the future.

Spoiler: I HAVE NO CLUE OF WHAT IS GOING ON, and I've been trying to understand this issue for some time.

# How to make it break

At the time of writing, this is (almost) the tests of archive compressing and decompressing ([click here to jump to the code](https://github.com/ouch-org/ouch/blob/1a67da03ed4b747e89619b0060082046019a35cf/tests/compress_and_decompress.rs#L10-L31)):
```rust
#[test]
/// Tests each format that supports multiple files with random input.
fn test_each_format() {
    test_compressing_and_decompressing_archive("tar");
    test_compressing_and_decompressing_archive("tar.gz");
    test_compressing_and_decompressing_archive("tar.bz");
    test_compressing_and_decompressing_archive("tar.bz2");
    test_compressing_and_decompressing_archive("tar.xz");
    test_compressing_and_decompressing_archive("tar.lz");
    test_compressing_and_decompressing_archive("tar.lzma");
    test_compressing_and_decompressing_archive("zip");
    test_compressing_and_decompressing_archive("zip.gz");
    test_compressing_and_decompressing_archive("zip.bz");
    test_compressing_and_decompressing_archive("zip.bz2");
    test_compressing_and_decompressing_archive("zip.xz");
    test_compressing_and_decompressing_archive("zip.lz");
    test_compressing_and_decompressing_archive("zip.lzma");
}
```

It works, but let's say I try to split some of these into different test functions.

```rust
#[test]
fn test_zip() {
    test_compressing_and_decompressing_archive("zip");
    test_compressing_and_decompressing_archive("zip.gz");
    test_compressing_and_decompressing_archive("zip.bz");
    test_compressing_and_decompressing_archive("zip.bz2");
    test_compressing_and_decompressing_archive("zip.xz");
    test_compressing_and_decompressing_archive("zip.lz");
    test_compressing_and_decompressing_archive("zip.lzma");
}

#[test]
fn test_tar() {
    test_compressing_and_decompressing_archive("tar");
    test_compressing_and_decompressing_archive("tar.gz");
    test_compressing_and_decompressing_archive("tar.bz");
    test_compressing_and_decompressing_archive("tar.bz2");
    test_compressing_and_decompressing_archive("tar.xz");
    test_compressing_and_decompressing_archive("tar.lz");
    test_compressing_and_decompressing_archive("tar.lzma");
}
```

Or even one test per `test_compressing_and_decompressing_archive` call.

## What breaks?

When we run more than one test at once, a lot of stuff break, here are some symptoms:
- A lot of "File or directory not found"
  - Files passed for compression disappear before opening (including opening for .zip archive building)
  - Decompressed files may be disappearing too
- Compared file contents after compressing and decompressing suddenly mismatch

## My suspections:

I suspect it is caused because different tests are run in different threads.

Ouch is _100% unsafe free_, and I can't find where we could be causing this problem, so I do suspect it has something to do with some library we are using, even thought it is very unlikely.

- Maybe `tempfile::TempDir` is messing up with drop in different threads?
- `rand::SmallRng` is very unlikely to have an non-documented bug like this.
- Any compression crate, like `flate2`, `bzip2`, `tar`, `xz2` or `zip`?

figsoda:
rewrite most tests with [proptest](https://lib.rs/proptest), tests single file {,de}compression and directory {,de}compression

Also found an issue with chaining lz4 compressions: `tar.lz4.lz4` files are not being compressed correctly, so it is ignored for now

I don't think file permissions are working on windows and for zip files, so I disabled checking them in these situations 

Closes #30
Closes #31
Closes #50


figsoda:
Not sure why tests are failing on windows

marcospb19:
Will give a proper review in a couple days. Got a bit busy again.

figsoda:
resolved merge conflicts

tests/mime.rs
```diff
@@ -0,0 +1,55 @@
...
+        "application/x-xz",
+        "application/x-xz",
+        "application/x-xz",
+        "application/zstd",
+        "application/x-bzip2",
+        "application/x-bzip2",
+        "application/x-xz",
+        "application/x-xz",
+        "application/zstd",
+    ];
```
GabrielSimonetto, line 40: I've checked that if some of these breaks, the error will be descriptive enough as to point out exactly which case failed
```
# If we change `application/x-tar` to `application/ERROR`
  left: `"application/x-tar"`,
 right: `"application/ERROR"`', tests/mime.rs:53:9
```

Still, I believe this test could become more descriptive if we used some sort of parametrizing strategy on this function. I know that the [test-case](https://github.com/frondeus/test-case) module can do this, but since we already are using proptest, does prop test introduce some sort of "Fixed parameters list" for this sort of functionality?

```
# How we could do it with test-case:
    #[test_case("tar",  "application/x-tar" )]
    #[test_case("zip", ""application/zip")]
    ...
    #[test_case("tar.zst",  "application/zstd")]
    fn sanity_check_through_mime(format: &str, expected_mime: &str) {
        ...
    }
```

I'll already extend this question to the other mantainers: If proptest can't do it, I'm fine with leaving this test as it is to avoid bringing `test-case` as a dev-dependency, do you agree?

tests/integration.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    // create more random files in 0 to 3 new directories
+    for _ in 0..rng.next_u32() % 4 {
+        create_random_files(&tempfile::tempdir_in(dir).unwrap().into_path(), depth - 1, rng);
+    }
+}
+
+// compress and decompress a single empty file
+#[proptest(cases = 512)]
+fn single_empty_file(ext: Extension, #[any(size_range(0..8).lift())] exts: Vec<FileExtension>) {
```
GabrielSimonetto, line 75: I took some time to understand how proptest works, and it seems like `ext: Extension` and `Vec<FileExtension>` can create generic examples for each enum, right? I really like this addittion, but I personally had some trouble in understanding the tests, since they show up as regular tests in the test suite:

```
     Running tests/integration.rs (target/debug/deps/integration-86976fb3c0298888)

running 5 tests
test utils::src_is_not_tests - should panic ... ok
test utils::src_is_src ... ok
test single_empty_file ... ok
test single_file ... ok
test multiple_files ... ok
```

Seeing at first glance, it is hard to understand that we are in fact checking a lot of stuff here, would you know of a way in which we could check which arguments were passed to this function, as to understand better how it is working?

If, for example, proptest could write what it used to a file, that would already improve a lot the understandability, something akin to this:

```
$ cargo test — proptest_write_stuff — tests.txt

$ cat tests.txt

# single_empty_file arguments:
ext = "zip", 3, exts = vec["zip", "gz" "bz"];
ext = "tar", 2, exts = vec["tar", "gz"];
...

# multiple_files arguments:
ext = "zip", exts = vec["zip", "gz" "bz"], depth = 3;
ext = "tar", exts = vec["tar", "gz"], depth = 2;
...

```

Of course this is only an example, anything you can think of would be helpful

GabrielSimonetto:
I really like this addition, but the scary part for me, as a non-initiated proptest user, is if we will be able to interpret the tests and be sure that they are doing their job along the time.

Also, we probably should write something about it on `CONTRIBUTING.md` as to how to work with it... there should be some stuff to write about on it, but the first thing that comes to mind is just a reference to the [proptest book](https://altsysrq.github.io/proptest-book/proptest/), but also, how we should treat the [proptest_regressions dir](https://altsysrq.github.io/proptest-book/proptest/failure-persistence.html), (which, the book seems to say that it should be added to git registry? idk).

I will keep looking into it so as to help understand how we should work with proptest, but you can help guide us on how to make this workflow more understandable that would be neat (:

tests/mime.rs
figsoda: I didn't do anything to this test except for removing the no longer needed sleep because I didn't know what to do about it
I think it would be a little confusing to use test_case and proptest at the same time, but we can probably implement a trait to convert `Extension` to mime types and integrate proptest with that

GabrielSimonetto:
Btw, have you understood how these changes fixed #50 ? It would be nice to know what we were doing wrong, for future reference

tests/integration.rs
figsoda: I don't know if there is an easy way to put all the arguments into a file, though the error messages generated by proptest are pretty readable imo

figsoda:
Added some more information to CONTRIBUTION.md

I have no idea how this fixes #50. I actually gave up investigating #50 and just bundled with a couple other issues and rewrote the tests.

tests/mime.rs
GabrielSimonetto: Yea, I agree with you that we don't need to do that right now, since it's not the focus of your PR and it was already this way beforehand.

vrmiguel (approved):
Very cool :D

I think that the `assert_cmd` approach is very interesting but it could bite us in the future if we ever decide to add code coverage metrics, I guess

tests/integration.rs
marcospb19: All we would need here are the error messages anyways, so resolving.

marcospb19:
> I have no idea how this fixes #50. I actually gave up investigating #50 and just bundled with a couple other issues and rewrote the tests.

That's a mystery, I wish anyone knew whats causing it.

tests/integration.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// extensions of single file compression formats
+#[derive(Arbitrary, Debug, Display)]
+#[display(style = "lowercase")]
+enum FileExtension {
+    Bz,
+    Bz2,
+    Gz,
+    // Lz4, // broken
```
marcospb19, line 36: Do you have any clues of what is causing this one? Maybe Lz4 would need to be filtered out in the next release?

figsoda: in normal cases its fine, but some formats are failing (`tar.lz4.lz4`, `tar.bz.lz4`)
It might have something to do with the library ([lz4_flex](https://lib.rs/crates/lz4_flex)), we can consider switching to liblz4 bindings like [lz4](https://lib.rs/crates/lz4) or [lzzzz](https://lib.rs/crates/lzzzz)

PSeitz: Can you give me something to reproduce? Compressing compressed data should work

marcospb19: @PSeitz if you prefer I can apply that on a branch for you to take a look at.

PSeitz: Thanks I already applied it and try to extract the data which is failing

PSeitz: I fixed the issues and created a PR here: https://github.com/ouch-org/ouch/pull/445

PR #172, opened 2021-11-07
Allow ignoring hidden files, respecting .gitignore and git exclude

### LINKED ISSUE #170: Add flag to skip files ignored by `.gitignore`
Hi there!
Very nice tool. I'm wondering, since this is using walkdir I'd assume adding the [ignore](https://crates.io/crates/ignore) crate ought to be fairly straight forward. So `ouch` could have a feature where it automatically ignores certain files/folders like `./.git` or `./node_modules`or `./target` while compressing.

Would anybody be interested in this functionality?

thank you

vrmiguel:
Closes #170

### DIFF
--- a/src/opts.rs
+++ b/src/opts.rs
@@ -42,6 +42,10 @@ pub enum Subcommand {
         /// The resulting file. Its extensions can be used to specify the compression formats.
         #[clap(required = true, value_hint = ValueHint::FilePath)]
         output: PathBuf,
+
+        /// If set, `ouch` will *not* compress hidden files or files ignored through .gitignore or .ignore
+        #[clap(short = 'H', long)]
+        hidden: bool,
     },
     /// Decompresses one or more files, optionally into another folder.
     #[clap(alias = "d")]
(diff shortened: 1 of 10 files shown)


AntonHermann:
This is awesome :D

One suggestion: I think splitting the options for git and hidden files would be valuable, maybe we could just copy `ripgrep`s options there?

vrmiguel:
> This is awesome :D
> 
> One suggestion: I think splitting the options for git and hidden files would be valuable, maybe we could just copy `ripgrep`s options there?

Good point! I'll try adding that soon :D

src/opts.rs
```diff
@@ -42,6 +42,10 @@ pub enum Subcommand {
         /// The resulting file. Its extensions can be used to specify the compression formats.
         #[clap(required = true, value_hint = ValueHint::FilePath)]
         output: PathBuf,
+
+        /// If set, `ouch` will *not* compress hidden files or files ignored through .gitignore or .ignore
```
marcospb19, line 46: This will appear in the `--help` message, so it should be a bit shorter, suggestion:

```rust
/// Ignore hidden files
```

I think that we should have another flag, `-i/--ignore` for `.gitignore` and `.ignore` files.

marcospb19:
Reopened and merged at #245.

Issue #131, opened 2021-10-30
Better error for could not create archive - append the failing output path


GabrielSimonetto:
While working on #94 we wound up wanting to build better errors for some cases, one of them is [this FinalError](https://github.com/ouch-org/ouch/blob/master/src/archive/tar.rs#L68):

```
builder.append_file(path, file.file_mut()).map_err(|err| {
    FinalError::with_title("Could not create archive")
        .detail("Unexpected error while trying to read file")
        .detail(format!("Error: {}.", err))
        .into_owned()
})?;
```

It isn't a great error message since it doesn't tell you which file broke, in a scenario where you are compressing/decompressing multiple files this could be an issue.

Now, it is troublesome to insert the output_file, since it was inserted on the tar [writer](https://github.com/ouch-org/ouch/blob/master/src/archive/tar.rs#L49) from it's [generic call](https://github.com/ouch-org/ouch/blob/master/src/commands.rs#L186-L198). This is a problem because:
* It seems that the Writer object does not have a getter to the output_path where it intends to write stuff, so we can't get this information from it.
* So we would end up trying to pass output_path on the function? Maybe, but since it's a generic call we would need to reproduce it to the zip call aswell I think.

So... I can't see a clear path of action, and this is already a hard error to occur aswell, therefore I will put the discussion tag and the low-priority one

marcospb19:
Just stopped to think about it, and we don't give this amount of context info when compressing with `zip` too.

And there are multiple points in that source code where errors can appear, so I guess we need to put custom error messages outside of both calls.

Maybe one message will do it for both, `Failed to create the archive {}`, fits for zip and tar.

What do you think?

GabrielSimonetto:
That sounds ideal, you're saying that if the error breaks inside tar or zip modules, it can propagate the error upwards and then the calling method can finish annotating context? I like that

PR #181, opened 2021-11-10
Improve zip errors when paths are not utf8 valid

marcospb19:


marcospb19:
```rust
if !invalid_unicode_filenames.is_empty() {
    let error = FinalError::with_title("Cannot build zip archive")
        .detail("Zip archives require files to have valid UTF-8 paths")
        .detail(format!("Files with invalid paths: {}", concatenate_os_str_list(&invalid_unicode_filenames)));

    return Err(error.into());
}
```

@GabrielSimonetto this is an error produced inside of the archive logic, but we don't have the name of the path that is being outputted, I just had an new idea!

What if we grab the Result<_, Error> and call a function on it before propagating up.

What this function would do: check if it is Error::Custom, and add a .detail() line to it, then returning with `?`. This way we could add the name of the `output_path` to an FinalError that was created INSIDE of the archive functions.

This idea is meant to solve #131.

GabrielSimonetto:
Hmm, but I feel like we need to go back to the caller to know which was the output_path, but you said:

> What if we grab the Result<_, Error> and call a function on it *before* propagating up.

How will a function inside this context be able to grab the output_path?

marcospb19:
> How will a function inside this context be able to grab the output_path?

Would be done with a function outside.

Inside of the `zip` function:

```rust
if !invalid_unicode_filenames.is_empty() {
    let error = FinalError::with_title("Cannot build zip archive")
        .detail("Zip archives require files to have valid UTF-8 paths")
        .detail(format!("Files with invalid paths: {}", concatenate_os_str_list(&invalid_unicode_filenames)));

    return Err(error.into());
}
```

Outside currently:

```rust
archive::zip::build_archive_from_paths(&files, &mut vec_buffer)?;
```

But could be like:

```rust
archive::zip::build_archive_from_paths(&files, &mut vec_buffer)
    .map_err(|err| add_detail(err, format!("Occurred in {}", output_path.display())));
```

marcospb19:
But I'm convinced that we don't really need this for compression, because there will always just be one compression output file.

But for extracting, we need to say the name.

PR #187, opened 2021-11-11
put compression backends behind features, clean up Cargo.toml

figsoda:
this way it is possible to change compression backends using `--no-default-features`

Also default zip to deflate-zlib backend since the same feature is already enabled by default in the flate2 dependency

I can add the alignment back if you want to, I just personally think it is better

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,20 +13,20 @@ description = "A command-line utility for easily compressing and decompressing f
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-clap        = "=3.0.0-beta.5" # Keep it pinned while in beta!
-atty        = "0.2.14"
-fs-err      = "2.6.0"
-once_cell   = "1.8.0"
-walkdir     = "2.3.2"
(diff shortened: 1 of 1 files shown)


SpyrosRoum:
Personally I don't really think this is necessary or needed. The whole point of ouch is to unite all these features, so unless one of them stands out (and has some runtime dependencies, or works differently for example) then I don't see a need for this.
But this is just my opinion of course

marcospb19:
But @SpyrosRoum, nothing has really changed, in theory this just added the capability for people to switch features if they want to.

image

Output of `ldd` is the same too:

image

(Had to run `sed` to remove memory address "`(0x00007f4cafb1b000)`" parts cause branches differ a bit).

SpyrosRoum:
I know in practice it doesn't really make a difference, I don't have a strong opinion on this either way

marcospb19:
I mean, I agree it's not really necessary :thinking: , but I don't plan on adding this to our CI anyway, if someone want to compile with different features, they'll be on their own, so we won't have any additional trouble.

It also unlikely that people will be using this, but because I see no downsides, I'll be merging.

PR #188, opened 2021-11-11
remove trailing blank lines in error messages

### LINKED ISSUE #184: Remove trailing blank line at error messages
Example:

image

It should not have that blank line.

figsoda:
fixes #184

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -49,11 +49,11 @@ pub struct FinalError {
 impl Display for FinalError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         // Title
-        writeln!(f, "{}[ERROR]{} {}", *RED, *RESET, self.title)?;
+        write!(f, "{}[ERROR]{} {}", *RED, *RESET, self.title)?;
 
         // Details
         for detail in &self.details {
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -49,23 +49,23 @@ pub struct FinalError {
 impl Display for FinalError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         // Title
-        writeln!(f, "{}[ERROR]{} {}", *RED, *RESET, self.title)?;
+        write!(f, "{}[ERROR]{} {}", *RED, *RESET, self.title)?;
 
         // Details
         for detail in &self.details {
-            writeln!(f, " {}-{} {}", *WHITE, *YELLOW, detail)?;
+            write!(f, "\n - {}{}{}", *YELLOW, detail, *RESET)?;
```
marcospb19, line 56: FYI, just in case you don't know it already, `WHITE != RESET` but not in all environments.

marcospb19 (approved):

src/error.rs
figsoda: imo `RESET` is better here for terminals with a light background

marcospb19: Ooooooh trueeeeee, agreed, I have never even thought about this before, always be using dark theme.

PR #190, opened 2021-11-11
Fix not overwriting files/dirs when trying to create a dir

### LINKED ISSUE #136: Decompression error when folder already exists at the place of output file

SpyrosRoum:
Closes #136 hopefully for good

### DIFF
--- a/src/utils/fs.rs
+++ b/src/utils/fs.rs
@@ -19,14 +19,14 @@ pub fn dir_is_empty(dir_path: &Path) -> bool {
     dir_path.read_dir().map(is_empty).unwrap_or_default()
 }
 
-/// Remove `path` asking the user to overwrite if necessary
-/// `Ok(Some(())` means the path is clear,
-/// `Ok(None)` means the user doesn't want to overwrite
-/// `Err(_)` is an error
-// ToDo: Actual type to translate the above might be clearer?
-pub fn clear_path(path: &Path, question_policy: QuestionPolicy) -> crate::Result<Option<()>> {
+/// Remove `path` asking the user to overwrite if necessary.
+///
+/// * `Ok(true)` means the path is clear,
+/// * `Ok(false)` means the user doesn't want to overwrite
+/// * `Err(_)` is an error
+pub fn clear_path(path: &Path, question_policy: QuestionPolicy) -> crate::Result<bool> {
     if path.exists() && !user_wants_to_overwrite(path, question_policy)? {
-        return Ok(None);
+        return Ok(false);
     }
 
     if path.is_dir() {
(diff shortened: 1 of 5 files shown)


src/utils/fs.rs
```diff
@@ -19,6 +19,25 @@ pub fn dir_is_empty(dir_path: &Path) -> bool {
     dir_path.read_dir().map(is_empty).unwrap_or_default()
 }
 
+/// Remove `path` asking the user to overwrite if necessary
+/// `Ok(Some(())` means the path is clear,
+/// `Ok(None)` means the user doesn't want to overwrite
+/// `Err(_)` is an error
+// ToDo: Actual type to translate the above might be clearer?
```
marcospb19, line 26: What if you change:
```
Some() -> true
None -> false
```

Using `Result<bool>`

And add to the docs that "Returns whether the path is clear or not"

marcospb19 (changes requested):
Brilliant! Will just be addressing your `ToDo` comment:

src/utils/fs.rs
SpyrosRoum: Oh yeah, not sure why I didn't think of that.

marcospb19 (approved):

sigmaSd:
This line https://github.com/ouch-org/ouch/pull/190/files#diff-b86d43fce6a9c57160aa925719550d40a6ed3f5e5a0117679905f76e68acc520R345 removes the current directory  if the user doesn't use the --dir flag (because outputdir = ".")

and the message isn't obvious either `Do you want to overwrite ''? [Y/n] `

marcospb19:
If you turn `strip_cur_dir` into this:

```rust
pub fn strip_cur_dir(source_path: &Path) -> &Path {
    source_path
}
```

So it does not affect the code anymore, then.

```
Do you want to overwrite ''? [Y/n]
```

Turns into
```
Do you want to overwrite '.'? [Y/n]
```

Which proves the point that, indeed, it's trying to overwrite the current directory, I ran into this earlier today (at 3 AM).

sigmaSd:
Personally I prefer if in that message the path was canonicalized, was this discussed before?

marcospb19:
Kinda, canonicalize paths are something I was with intent of fixing it myself, but I'd prefer 'current directory' instead of the full path, but... this should never appear in this error message anyways.

I think that this
```
Do you want to overwrite 'LICENSE'? [Y/n]
```
Is better than
```
Do you want to overwrite '/home/sigmaSd/ouch/LICENSE'? [Y/n]
```

PR #189, opened 2021-11-11
add supported formats to help message

figsoda:
closes #185

PR #191, opened 2021-11-11
add link to github to help message

figsoda:
Follow up to #189

### DIFF
--- a/src/opts.rs
+++ b/src/opts.rs
@@ -6,6 +6,8 @@ use clap::{Parser, ValueHint};
 /// A command-line utility for easily compressing and decompressing files and directories.
 ///
 /// Supported formats: tar, zip, bz/bz2, gz, lz4, xz/lz/lzma, zst
+///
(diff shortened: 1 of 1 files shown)


src/opts.rs
```diff
@@ -6,6 +6,8 @@ use clap::{Parser, ValueHint};
 /// A command-line utility for easily compressing and decompressing files and directories.
 ///
 /// Supported formats: tar, zip, bz/bz2, gz, lz4, xz/lz/lzma, zst
```
marcospb19, line 8:
```suggestion
/// Supported formats: tar, zip, bz/bz2, gz, lz4, xz/lz/lzma, zst.
```

marcospb19 (approved):

marcospb19:
Thanks!

src/opts.rs
marcospb19: Commited this period but not in the URL because it breaks how some terminals deal with URL links

Issue #180, opened 2021-11-10
Unspecified behavior when compressing overwritten file


marcospb19:
You build the archive:
```
ouch compress * archive.tar.gz
```

If you go there and run it again:
```
ouch compress * archive.tar.gz
```

It will ask "do you wish to overwrite `archive.tar.gz`?

If you press `Y`, it will delete the file, open a new file for `archive.tar.gz`, but note that the same file was also passed as an input for this compression! (with `*` wildcard)

The same file is the input and output.

`ouch` will read the output file to add as input to the... same output file.

Also:

```
ouch compress LICENSE a.zip
```
Then
```
ouch compress a.zip a.zip a.zip a.zip
```

---

Maybe just adding a `FinalError` for when the output file was passed as input too would solve it.

SpyrosRoum:
Crashing feels a little too much. 
Maybe a warning and skipping that file is enough, thoughts?

marcospb19:
I agree! That's a great idea actually, it's what `mv` does when you're trying to move the directory to inside of itself.

![image](https://user-images.githubusercontent.com/38900226/141316668-7f9517a1-eda3-4e48-a4ba-9c1a510ea972.png)

PR #193, opened 2021-11-12
Skip compressing file if its the same file as the output

sigmaSd:
Fix https://github.com/ouch-org/ouch/issues/180

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -44,7 +44,16 @@ fn represents_several_files(files: &[PathBuf]) -> bool {
 /// Entrypoint of ouch, receives cli options and matches Subcommand to decide what to do
 pub fn run(args: Opts, question_policy: QuestionPolicy) -> crate::Result<()> {
     match args.cmd {
-        Subcommand::Compress { files, output: output_path } => {
+        Subcommand::Compress { mut files, output: output_path } => {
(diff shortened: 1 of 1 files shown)


src/commands.rs
```diff
@@ -44,7 +44,16 @@ fn represents_several_files(files: &[PathBuf]) -> bool {
 /// Entrypoint of ouch, receives cli options and matches Subcommand to decide what to do
 pub fn run(args: Opts, question_policy: QuestionPolicy) -> crate::Result<()> {
     match args.cmd {
-        Subcommand::Compress { files, output: output_path } => {
+        Subcommand::Compress { mut files, output: output_path } => {
+            // If the output_path file exists and is the same as some of the input files, warn the user and skip those inputs (in order to avoid compression recursion)
+            if output_path.exists() {
+                clean_input_files_if_needed(&mut files, &output_path.canonicalize()?);
```
marcospb19, line 50: Here you'll need to call `fs::canonicalize()` instead of `Path::canonicalize` because we are using `fs_err` aliased to `fs`, but `Path::canonicalize` will be calling `std::fs::canonicalize` instead.

marcospb19 (changes requested):

src/commands.rs
sigmaSd: done

marcospb19 (approved):

Issue #151, opened 2021-11-02
Automatically decompress the archive into a directory if there is no directory for the contents in the archive


c02y:
For instance, I got a zip file containing multiple dirs and files in it, but they have no parent directory.

When you decompress it using `ouch d dir.zip` all the contents will be decompressed into the current directory which may already contains a lot of other dirs or files.

It is better to automatically decompress the archive into a new directory, the name of the directory should be the file name of the archive without the extension. If you think it is not a good idea by default, an option can be provided.

For example, command `unar` can do it(`unar` is the command from `unarchiver` package, link is https://github.com/MacPaw/XADMaster)

GabrielSimonetto:
Yea, this probably should be the default config, and if someone wants to unpack inplace we can provide a flag for that

marcospb19:
This practically discards the necessity of the `ouch revert` command I thought about to clean up, reverting everything that was done in the last command.

sigmaSd:
I think ouch already does this for single file formats, so maybe a simple patch like this is enough to add this behavior for archive file formats:
 
```diff
diff --git a/src/commands.rs b/src/commands.rs
index 7550bae..8051de0 100644
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -338,7 +338,8 @@ fn decompress_file(
         if let Some(output_dir) = output_dir { output_dir.join(file_name) } else { file_name.to_path_buf() };
 
     // Output folder is used by archive file formats (zip and tar)
-    let output_dir = output_dir.unwrap_or_else(|| Path::new("."));
+    let owned_file_name = file_name.to_path_buf();
+    let output_dir = output_dir.unwrap_or_else(|| &owned_file_name);
 
     // Zip archives are special, because they require io::Seek, so it requires it's logic separated
     // from decoder chaining.
```

runiq:
> I think ouch already does this for single file formats, so maybe a simple patch like this is enough to add this behavior for archive file formats:
> 
> ```diff
> diff --git a/src/commands.rs b/src/commands.rs
> index 7550bae..8051de0 100644
> --- a/src/commands.rs
> +++ b/src/commands.rs
> @@ -338,7 +338,8 @@ fn decompress_file(
>          if let Some(output_dir) = output_dir { output_dir.join(file_name) } else { file_name.to_path_buf() };
>  
>      // Output folder is used by archive file formats (zip and tar)
> -    let output_dir = output_dir.unwrap_or_else(|| Path::new("."));
[...]

PR #195, opened 2021-11-12
[Draft] Smart decompress into a directory

sigmaSd:
fix https://github.com/ouch-org/ouch/issues/151

This a bit complex and I managed to remove directories unintentionally a couple of times when writing it, so the maintainers needs to judge if this complexity is worth it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ walkdir = "2.3.2"
 xz2 = "0.1.6"
 zip = { version = "0.5.13", default-features = false }
 zstd = { version = "0.9.0", default-features = false }
+tempfile = "3.2.0"
 
 [build-dependencies]
 clap = "=3.0.0-beta.5"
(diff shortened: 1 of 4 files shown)


marcospb19:
Do you have idea why it's failing on `Windows`? Maybe some tempfile differences stuff?

sigmaSd:
I added an XXX note where the problem is

Rename cant happen between different mount points (like /tmp and /)

Thats why in proptest the tempdir should be created in /tmp but somehow proptest still takes the tempdir_in(".") path and create it in the wrong place

Im not sure whats the real issue I thought proptest should invoke the code under cfg test but it isnt, maybe its a bug

sigmaSd:
Thats whats hapenning on linux, Im guessing its the same issue in windows

sigmaSd:
Maybe a more correct approach is to create the temp directory in the parent folder of the output file instead of hardcoding to "."

I think that would make it work im all cases

marcospb19:
Got it, so, I did not had the time to review this properly yet, `ouch` has been getting a lot of attention and it's hard to keep up.

sigmaSd:
That's a good thing!

sigmaSd:
I'm not sure about the windows test yet

Also there are still some edge cases not handled

sigmaSd:
Superseded by https://github.com/ouch-org/ouch/pull/209

PR #196, opened 2021-11-12
Fix warnings in doc comments

AntonHermann:
Add automatic link markup to urls

Escape square brackets comments containing "[INFO]", "[WARNING]", "[ERROR]"

### DIFF
--- a/src/utils/formatting.rs
+++ b/src/utils/formatting.rs
@@ -9,7 +9,7 @@ use std::{
 ///
 /// This is different from [`Path::display`].
 ///
-/// See https://gist.github.com/marcospb19/ebce5572be26397cf08bbd0fd3b65ac1 for a comparison.
+/// See <https://gist.github.com/marcospb19/ebce5572be26397cf08bbd0fd3b65ac1> for a comparison.
 pub fn to_utf(os_str: impl AsRef<OsStr>) -> String {
     let text = format!("{:?}", os_str.as_ref());
     text.trim_matches('"').to_string()
--- a/src/utils/fs.rs
+++ b/src/utils/fs.rs
@@ -59,7 +59,7 @@ pub fn cd_into_same_dir_as(filename: &Path) -> crate::Result<PathBuf> {
 }
 
 /// Try to detect the file extension by looking for known magic strings
-/// Source: https://en.wikipedia.org/wiki/List_of_file_signatures
+/// Source: <https://en.wikipedia.org/wiki/List_of_file_signatures>
 pub fn try_infer_extension(path: &Path) -> Option<Extension> {
     fn is_zip(buf: &[u8]) -> bool {
         buf.len() >= 3
--- a/src/error.rs
+++ b/src/error.rs
@@ -38,7 +38,7 @@ pub type Result<T> = std::result::Result<T, Error>;
 /// Pretty final error message for end users, crashing the program after display.
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct FinalError {
-    /// Should be made of just one line, appears after the "[ERROR]" part
+    /// Should be made of just one line, appears after the "\[ERROR\]" part
     title: String,
     /// Shown as a unnumbered list in yellow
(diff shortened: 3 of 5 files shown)


src/opts.rs
```diff
@@ -7,7 +7,7 @@ use clap::{Parser, ValueHint};
 ///
 /// Supported formats: tar, zip, bz/bz2, gz, lz4, xz/lz/lzma, zst.
 ///
-/// Repository: https://github.com/ouch-org/ouch
+/// Repository: <https://github.com/ouch-org/ouch>
```
AntonHermann, line 10: Because this doc comment is used in both the `clap`-generated help output and the `rustdoc` generated documentation, this results in the angle brackets showing up in the `clap` output as well.

In my opinion this is something that would be nice to get fixed by `clap`, so I started an issue over there as well: [clap-rs/clap issue 3019](https://github.com/clap-rs/clap/issues/3019#issue-1052332655) although it doesn't have high priority there (totally understandable)

If having these angle brackets in the help output bothers you, we could try to duplicate this doc comment, using the doc comment itself for `rustdoc` and providing the long help for clap via `#[clap(long_about = "")]`

figsoda: > his results in the angle brackets showing up in the clap output as well.

I don't think this is worth it to remove rustdoc warnings since it isn't really used as a library and pretty much everyone use the help function

> If having these angle brackets in the help output bothers you, we could try to duplicate this doc comment, using the doc comment itself for rustdoc and providing the long help for clap via #[clap(long_about = "")]

maybe we can also just try to separate them and not have rustdoc touch any of it if it is possible. I don't see a reason this would be useful in a library anyways. Duplicating text isn't really an elegant solution

AntonHermann: Yeah you're probably right. I'll revert the URL changes and add an `#![allow(rustdoc::bare_urls)]` in main.rs

AntonHermann: I hope it's okay that I force-pushed this change, didn't want to have 2 separate commits for such a tiny change

src/utils/fs.rs
```diff
@@ -59,7 +59,7 @@ pub fn cd_into_same_dir_as(filename: &Path) -> crate::Result<PathBuf> {
 }
 
 /// Try to detect the file extension by looking for known magic strings
-/// Source: https://en.wikipedia.org/wiki/List_of_file_signatures
+/// Source: <https://en.wikipedia.org/wiki/List_of_file_signatures>
```
marcospb19, line 62: Did you forgot this one?

src/utils/formatting.rs
```diff
@@ -9,7 +9,7 @@ use std::{
 ///
 /// This is different from [`Path::display`].
 ///
-/// See https://gist.github.com/marcospb19/ebce5572be26397cf08bbd0fd3b65ac1 for a comparison.
+/// See <https://gist.github.com/marcospb19/ebce5572be26397cf08bbd0fd3b65ac1> for a comparison.
```
marcospb19, line 12: And this.

src/utils/fs.rs
AntonHermann: No, only the doc comments in _src/opts.rs_ are affected since they are the only ones that get used by clippy to generate help output. This and the URL in _src/utils/formatting.rs_ are purely internal documentation, therefore we can make use of the rustdoc automatic link markup here :)

marcospb19: Oh yeah xD my bad.

AntonHermann:
Regarding the URL markup not getting correctly parsed by `clap`, which is why we don't use `rustdoc`s automatic link markup on user-facing doc comments (those that are used by clap to generate CLI help texts):

The clap issue I mentioned earlier was closed because there already is clap-rs/clap#2389 addressing a superset of this automatic link markup (i.e. parsing markdown/markup in general). If this is fixed someday, we could add the markup in _src/opts.rs_ again (this has _suuper low importance_ :D

PR #197, opened 2021-11-13
Implement accessibility mode

### LINKED ISSUE #194: Make ouch accessible
The warnings and info messages are not accessible at the moment, the square brackets are most likely read out as square brackets and I'm not sure whether having a message for each (de)compressed file is valuable for a sight-impaired person.

**Motivation**
See [this blogpost](https://bees.substack.com/p/making-hacking-accessible) for some arguments/guidelines for making hacking accessible (there are many CLI specific informations)

**What should change?**
Not much actually, I think adding an `accessibility` option and disabling all `info!` output as well as removing the square brackets for `warning!` output and checking all `(e)print(ln)!`s whether they provide value to a sight-impaired person.

**Implementation ideas**
I don't know whether it would be too much clutter to pass this option to every output macro call, so here are some ideas how this could be implemented in a less noise way:
- Pass it to every output macro, but extract this and other flags into a new struct that can be passed around comfortably (the subset of `opts::Opts` that can be `Copy`ed)
- Use an env var instead which can be checked in the output macros
- global state (using `once_cell`)

_This could be advertised as a feature in the readme. Many CLI tools are not designed with accessibility in mind. We could make a difference towards making CLI more accessible in general and thus including an often forgotten minority! 🎉_

AntonHermann:
Fixes #194 

This is my take at implementing an accessible command line interface.

When either setting the environment variable `ACCESSIBLE` (to an arbitrary value) or passing the CLI flag `--accessible`/`-A`, execute in accessibility mode: reduce visual noise and unnecessary output (almost all `info!()` calls)

When setting the environment variable `ACCESSIBLE` to a nonempty value when executing _install.sh_, suppress noisy output of `curl`/`wget`

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,12 +24,13 @@
 # Features
 
 1. Easy to use.
-2. Automatic formats detection.
-3. Same usage syntax for all formats.
-4. Uses encoding and decoding streams to improve performance.
-5. No runtime dependencies (for _Linux x86_64_).
-6. Can list archive contents with pretty tree formatting.
-7. Shell completions (soon!).
+2. Accessibility (A11Y) mode via `--accessibility` or env var `ACCESSIBILITY`
+3. Automatic formats detection.
+4. Same usage syntax for all formats.
+5. Uses encoding and decoding streams to improve performance.
+6. No runtime dependencies (for _Linux x86_64_).
+7. Can list archive contents with pretty tree formatting.
+8. Shell completions (soon!).
 
 # Usage
 
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@
 # Features
 
 1. Easy to use.
-2. Accessibility (A11Y) mode via `--accessibility` or env var `ACCESSIBILITY`
+2. Accessibility mode (A11Y) via `--accessibility` or `ACCESSIBILITY` env var (see [wiki page](https://github.com/ouch-org/ouch/wiki/Accessibility)).
 3. Automatic formats detection.
 4. Same usage syntax for all formats.
 5. Uses encoding and decoding streams to improve performance.
--- a/install.sh
+++ b/install.sh
@@ -7,6 +7,10 @@ DOWNLOAD_LOCATION="/tmp/ouch-binary"
 INSTALLATION_LOCATION="/usr/local/bin/ouch"
 REPO_URL="https://github.com/ouch-org/ouch"
 
+# If env var ACCESSIBLE is set (to a nonempty value), suppress output of
+# `curl` or `wget`
+
+
 # Panics script if anything fails
 set -e
 
@@ -55,12 +59,15 @@ install() {
     echo ""
 
     # Set $downloader
+    downloader_quiet_flag=""
     if [ $(which curl) ]; then
         downloader="curl"
-        downloader_command="curl -fSL $binary_url -o $DOWNLOAD_LOCATION"
+        if [ "$ACCESSIBLE" ]; then downloader_quiet_flag="--silent"; fi
+        downloader_command="curl $downloader_quiet_flag -fSL $binary_url -o $DOWNLOAD_LOCATION"
     elif [ $(which wget) ]; then
         downloader="wget"
-        downloader_command="wget $binary_url -O $DOWNLOAD_LOCATION"
+        if [ "$ACCESSIBLE" ]; then downloader_quiet_flag="--quiet"; fi
+        downloader_command="wget $downloader_quiet_flag $binary_url -O $DOWNLOAD_LOCATION"
     else
         echo "ERROR: have not found 'curl' nor 'wget' to donwload ouch binary."
         exit 1
--- a/src/error.rs
+++ b/src/error.rs
@@ -49,7 +49,11 @@ pub struct FinalError {
 impl Display for FinalError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         // Title
-        write!(f, "{}[ERROR]{} {}", *RED, *RESET, self.title)?;
+        if *crate::cli::ACCESSIBLE.get().unwrap_or(&false) {
+            write!(f, "{}ERROR{}: {}", *RED, *RESET, self.title)?;
+        } else {
+            write!(f, "{}[ERROR]{} {}", *RED, *RESET, self.title)?;
+        }
 
         // Details
         for detail in &self.details {
--- a/src/error.rs
+++ b/src/error.rs
@@ -49,6 +49,8 @@ pub struct FinalError {
 impl Display for FinalError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         // Title
+        //
+        // When in ACCESSIBLE mode, the square brackets are suppressed
         if *crate::cli::ACCESSIBLE.get().unwrap_or(&false) {
             write!(f, "{}ERROR{}: {}", *RED, *RESET, self.title)?;
         } else {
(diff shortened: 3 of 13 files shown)


src/list.rs
```diff
@@ -23,7 +23,7 @@ pub struct FileInArchive {
 
 /// Actually print the files
 pub fn list_files(archive: &Path, files: Vec<FileInArchive>, list_options: ListOptions) {
-    println!("{}:", archive.display());
+    println!("Archiv: {}", archive.display());
```
marcospb19, line 26: Is Archiv a typo?

README.md
```diff
@@ -24,12 +24,13 @@
 # Features
 
 1. Easy to use.
-2. Automatic formats detection.
-3. Same usage syntax for all formats.
-4. Uses encoding and decoding streams to improve performance.
-5. No runtime dependencies (for _Linux x86_64_).
-6. Can list archive contents with pretty tree formatting.
-7. Shell completions (soon!).
+2. Accessibility (A11Y) mode via `--accessibility` or env var `ACCESSIBILITY`
```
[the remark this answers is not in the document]
AntonHermann: So in the "Features" list we just note that there is an accessibility mode and link a new wiki page for that?

I like the idea of a separate wiki page.

I've been reading through a [case study: "Accessibility of Command Line Interfaces "](https://dl.acm.org/doi/fullHtml/10.1145/3411764.3445544) and one of their recommendations was to provide a detailed description of the tool output since visually impaired people can't parse the structure of the output easily. Such a thing could be described over there as well.

src/list.rs
AntonHermann: yeah, that should have been "Archive: " :D

README.md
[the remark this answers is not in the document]
marcospb19: Something like
```
2. Accessibility mode (A11Y) via `--accessibility` or `ACCESSIBILITY` env var (see [wiki page](https://github.com/ouch-org/ouch/wiki/Accessibility)).
```

2. Accessibility mode (A11Y) via `--accessibility` or `ACCESSIBILITY` env var (see [wiki page](https://github.com/ouch-org/ouch/wiki/Accessibility)).

install.sh
```diff
@@ -55,12 +59,15 @@ install() {
     echo ""
 
     # Set $downloader
+    downloader_quiet_flag=""
     if [ $(which curl) ]; then
         downloader="curl"
-        downloader_command="curl -fSL $binary_url -o $DOWNLOAD_LOCATION"
+        if [ "$ACCESSIBLE" ]; then downloader_quiet_flag="--silent"; fi
```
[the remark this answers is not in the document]
AntonHermann: This isn't a thing as far as I know, I looked for guidelines and how other tools do it, but couldn't find anything. This seemed to be a sensible thing and if it's documented in the README and/or an accessibility wiki page I think affected users should be able to learn about it before trying to install and use `ouch`

AntonHermann:
Sorry about the long delay, I've been pretty busy recently.

This PR contains a fixup commit now which doesn't add any value of its own and there is a small conflict with master. How do you want to handle this?

I could create a merge commit, but I think squashing and rebasing all commits onto master would be cleaner. I would await another review and then squash, rebase and force-push this PR? This way we only have a single, significant commit being merged into master.

src/macros.rs
```diff
@@ -1,12 +1,26 @@
...
 /// Macro that prints \[INFO\] messages, wraps [`println`].
+///
+/// Normally, this prints nothing if ACCESSIBLE mode is turned on,
+/// except when called as `info!(a11y_show, "..", ..)`
 #[macro_export]
 macro_rules! info {
-    ($($arg:tt)*) => {
-        $crate::macros::_info_helper();
+    // Show info message even in ACCESSIBLE mode
+    (a11y_show, $($arg:tt)*) => {
```
AntonHermann, line 10: I'm not that happy with this `info!(a11y_show, ...)` and it confused others as well. The idea is to somehow mark some `info!()` calls as relevant for accessibility-mode output whereas all other `info!()` calls without `a11y_show` are suppressed. I'm struggling to come up with a better way to handle this, do you have any ideas what we could do here or whether this is okay as long as it's properly documented?

install.sh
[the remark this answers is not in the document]
marcospb19: That's a great point!

marcospb19:
> I could create a merge commit, but I think squashing and rebasing all commits onto master would be cleaner.

Whatever you'd prefer! So far, I did not spent much time on making git history clear, I know it's good, but sometimes I think that we are just low on time, and should pick the "less effort" alternative.

About the conflicts here, I'll solve them once the PR is ready, do not worry.

AntonHermann:
Okay great.
I'd like to clear up my doubts about https://github.com/ouch-org/ouch/pull/197#discussion_r753883576, then I will clean up the PR :)

src/error.rs
```diff
@@ -49,7 +49,11 @@ pub struct FinalError {
 impl Display for FinalError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         // Title
-        write!(f, "{}[ERROR]{} {}", *RED, *RESET, self.title)?;
+        if *crate::cli::ACCESSIBLE.get().unwrap_or(&false) {
+            write!(f, "{}ERROR{}: {}", *RED, *RESET, self.title)?;
+        } else {
+            write!(f, "{}[ERROR]{} {}", *RED, *RESET, self.title)?;
+        }
```
marcospb19, line 56: You changed the title, but forgot to change the hints, I think.

Try running this:

```sh
$ touch input
$ ouch compress input output
```

image

---

Or this:

```sh
$ touch a
$ ouch compress a a.gz.tar
```

image

marcospb19 (changes requested):

src/macros.rs
marcospb19: Oops, saw this earlier today, but forgot to comment.

I also think thought at first that the macro is a bit confusing but I didn't knew how to improve it, so I just accepted it.

Now that you asked again, I think I came with an idea.

What about having:

```rust
info!(always, "Message here");
```

And:

```rust
info!(not_tts, "Message here");
```

The `always` one is always displayed, and the `not_tts` is disabled when the user might be using `tts`.

So then the person who is using `info!` will always have to keep in mind "is this message tts-friendly?

The name does not seem right, so here's another idea:

```rust
info!(accessible, "Message Here");
info!(inaccessible, "Message Here");
```

Accessible messages can be displayed every time, inaccessible messages are flood (too much text on screen).

We could have another macro, but, this will certainly lead to people not worrying about accessibility, forgetting to use the correct one.

Additionally, for this to cause less confusion, we'd need explanation in the docs on why some messages are hidden, not everyone knows what TTS tools are, and those people won't figure out that it's because too much text makes it hard to navigate and understand the output (if I'm not mistaken, that's the reason why we are using this macro versions).

AntonHermann: Neat, I think I prefer the second idea with _accessible_ and _inaccessible_. I think the term TTS may be unfamiliar to many devs..

AntonHermann: I'm almost done, just one `info!` remains that I'm unsure about:
https://github.com/ouch-org/ouch/blob/cb6d65c79aa917e367c994482c760682e97d8fbb/src/archive/zip.rs#L152
I am not very familiar with ZIP files and neither the function declaration nor the call in L.47 has any documentation/comments. Is this an important message?

Apart from that, I don't understand this even from a seeing person's viewpoint. What does it mean? What is a ZIP archive comment, why is this message printed?

src/error.rs
AntonHermann: Good catch!

I've changed to output to this for accessibility mode:
2021-11-23_11-47-01_Tuesday

marcospb19: Perfect!

src/macros.rs
marcospb19: Someone can insert a "comment" in the metadata of this zip file, telling where it comes from, or what it is, or a link, a name, anything.

Is one of the features of zip archives, like passwords (that we don't support yet because ... ).

I think we should display it, if a person inserts a book in the comment, our tool will not be accessible anymore, but it can be bad to hide this info.

Googled for "zip comment size limit":

> The .ZIP file format allows for a comment containing up to 65,535 (216−1) bytes of data to occur at the end of the file after the central directory.

It might contain thousands of words, or just one very useful piece of information.

I never see a comment in zip archives, so I don't have a very strong opinion on this, you can decide.

It's a tradeoff: show a possibly useful information or making sure our application is 100% accessible.

Both are reasonable choices.

AntonHermann: Neat, thanks for looking that up.

I marked it as accessible for now and included a comment stating that if we stumble across very long, unnecessary comments in the future, it might be sensible to ask the user first before printing the whole comment (showing the length of the comment beforehand).

But until then, I think just printing it is good enough

PR #198, opened 2021-11-13
Remove Lzip because its incorrect, and improve extension comparison

sigmaSd:
This pr does 2 things:
- Remove `lzip` (lz) because its was considered by ouch to be `lzma` but according to https://en.wikipedia.org/wiki/List_of_file_signatures its actually `lzip`
- Ignore display strings when comparing extensions, this especially useful for inferring because without it file that uses algorithms that can have different extensions wont be detected as equal (ex: lzma)


sigmaSd:
lzip needs https://www.nongnu.org/lzip/lzlib.html

The sys-crate can easily be generated with bindgen, but I guess we would need a nice rusty wrapper.

src/utils/fs.rs
```diff
@@ -110,10 +107,8 @@ pub fn try_infer_extension(path: &Path) -> Option<Extension> {
         Some(Extension::new(&[Gzip], "gz"))
     } else if is_bz2(&buf) {
         Some(Extension::new(&[Bzip], "bz2"))
-    } else if is_xz(&buf) {
-        Some(Extension::new(&[Lzma], "xz"))
-    } else if is_lz(&buf) {
-        Some(Extension::new(&[Lzma], "lz"))
+    } else if is_lzma(&buf) {
+        Some(Extension::new(&[Lzma], "lzma")) // lzma can be `xz` or `lzma`
```
figsoda, line 111: `xz` extensions seem to be more popular, should we use `"xz"` instead?

sigmaSd: That makes sense

firasuke:
@sigmaSd I've opened #247 and I've been working on:
- the low level crate [lzip-sys](https://crates.io/crates/lzip-sys) which is mostly finished, the bindings were generated using bindgen then manually cleaned and modified a bit
- the higher level crate [lzip-rs](https://crates.io/crates/lzip) is still being worked on while imitating what `bzip2-rs` and `xz2-rs` are doing

Let me hear your thoughts on this!

sigmaSd:
Hello @firasuke thanks for working on this

Personally I have no experience with those libraries and I see that you have already made some good progress.

I think you would have better chances asking bzip-2 and xz-2 maintainer for guidance.

Otherwise nice work looking forward for it to be used in ouch

firasuke:
@sigmaSd Thanks!

Actually `lzip-sys` is done, all that's left is the higher level api `lzip-rs` which I also have no experience in implementing.

I also tried to get in contact with Alex (lead developer of `xz2-rs` and `bzip2-rs`) to no avail.

Perhaps someone with more experience in Rust can give us their thoughts on this.

PR #199, opened 2021-11-13
improve issue templates

figsoda:
Follow up to #186

templates updated:
- bug report (fancy yaml template, [rendered](https://github.com/figsoda/ouch/blob/issue-templates/.github/ISSUE_TEMPLATE/bug-report.yml))
- documentation improvement (minor changes)

templates added:
- feature request
- question


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,40 @@
...
+    id: description
+    attributes:
+      label: Description
+    validations:
+      required: true
+
+  - type: textarea
+    id: actual
+    attributes:
+      label: Actual Output
```
marcospb19, line 27: I think it's better to call it `Current Behavior` and `Expected Behavior`, because some bugs are not directly related to "output".

.github/ISSUE_TEMPLATE/question.md
```diff
@@ -0,0 +1,6 @@
+---
+name: Question
+about: Ask a question about ouch
+labels: question
```
marcospb19, line 4: You should also throw some text to the question template description pointing to the [`Discussions`](https://github.com/ouch-org/ouch/discussions) tab, saying that questions can be asked there too.

marcospb19 (changes requested):

.github/ISSUE_TEMPLATE/question.md
figsoda: Should I redirect all questions to discussions tab and not have the template at all instead?

marcospb19: I'm not sure, I guess we can leave both channels open so people can choose depending on what is the question about.

marcospb19 (approved):
Thanks!

PR #206, opened 2021-11-14
Simplify decompress function

sigmaSd:
While working on https://github.com/ouch-org/ouch/pull/195 I felt that there is a lot of complexity in certain part of the code-base

This can be dangerous since ouch manipulates files and directories

This an attempt to simplify decompress function

One thing I would want is if we agree on some terminology and write it somewhere, for example the word `file` is used everywhere but sometimes it means directories, and I really want a name for single file format archive and the other type, etc

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -201,11 +201,16 @@ pub fn run(args: Opts, question_policy: QuestionPolicy) -> crate::Result<()> {
                 return Err(error.into());
             }
 
-            // From Option<PathBuf> to Option<&Path>
-            let output_dir = output_dir.as_ref().map(|path| path.as_ref());
+            // The directory that will contain the output files
+            // We default to the current directory if the user didn't specify an output directory with --dir
+            let output_dir = output_dir.unwrap_or_else(|| PathBuf::from("."));
+            // NOTE: If the user *did* use --dir and the directory already exists, we will merge the
+            // extracted filed inside it
+            utils::create_dir_if_non_existent(&output_dir)?;
 
             for ((input_path, formats), file_name) in files.iter().zip(formats).zip(output_paths) {
-                decompress_file(input_path, formats, output_dir, file_name, question_policy)?;
+                let output_file_path = output_dir.join(file_name); // Path used by single file format archives
+                decompress_file(input_path, formats, &output_dir, output_file_path, question_policy)?;
             }
         }
         Subcommand::List { archives: files, tree } => {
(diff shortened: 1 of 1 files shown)


src/commands.rs
```diff
@@ -342,11 +339,6 @@ fn decompress_file(
     //
     // Any other Zip decompression done can take up the whole RAM and freeze ouch.
     if formats.len() == 1 && *formats[0].compression_formats == [Zip] {
-        if !utils::clear_path(output_dir, question_policy)? {
-            // User doesn't want to overwrite
-            return Ok(());
-        }
-        utils::create_dir_if_non_existent(output_dir)?;
```
marcospb19, line 349: Why we won't clear path anymore?

marcospb19:
> One thing I would want is if we agree on some terminology and write it somewhere, for example the word file is used everywhere but sometimes it means directories, and I really want a name for single file format archive and the other type, etc

I mean, we accept files as input, but directories are files too.

The correct way to differentiate, [in my opinion](https://docs.rs/file_type_enum/0.11.1/file_type_enum/), is that a "file" can be a "regular file" or a "directory".

I used to call those formats "Single file formats", but you can call them "non-archive formats" too, although I think that "single file format archive" that you said does not make sense, they are not archives because archives have support to bundle multiple files inside of it.

I think of it this way: single file formats do not accept directories because they can contain multiple files inside.

src/commands.rs
sigmaSd: I hoisted that out of the function here https://github.com/ouch-org/ouch/pull/206/files/85968ea8410ea86b8a636aaacf1ff4280c3e0a84#diff-b86d43fce6a9c57160aa925719550d40a6ed3f5e5a0117679905f76e68acc520R209

There is a choice to make, what I went with here is if the directory exists we will just put the files in it without overwriting

The second choice is to ask for overwrite, but in that case we should *only* ask for overwrite if the user specified a dir with --dir (if he did not, the directory is "." and we should not ask to overwrite it)

Either way is fine, I just went with the 1st one because its less code

sigmaSd: I restored the previous behaviour

marcospb19: > Either way is fine, I just went with the 1st one because its less code

I guess I agree, I'm just a bit worried if this will change other stuff or add some bugs, because we had some new bugs recently in `ouch` because I was short on time and ended up merging a lot of PRs without testing everything.

sigmaSd: Yes I definitely want to add more tests, at least everyone will feel "safer" when adding prs

marcospb19: I don't think it's strictly necessary, majority of bugs I catch by running `ouch` two or four times but would be impossible to catch with tests, because they come in new features.

PR #208, opened 2021-11-14
Add an integration test

sigmaSd:
This is an experiment

The output in theory can be in-deterministic, (for example if  reading from ouch stdout reads only a few bytes and thus the assertion would fail), but I feel like this shouldn't happen in practice


marcospb19:
Had to reverted because this broke tests

PR #209, opened 2021-11-15
Add heuristics to decompressing archives

### LINKED ISSUE #151: Automatically decompress the archive into a directory if there is no directory for the contents in the archive

sigmaSd:
I fell way better about this PR, turns out that simplifying stuff is a correct step

fix #151

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -345,6 +345,7 @@ fn decompress_file(
     output_file_path: PathBuf,
     question_policy: QuestionPolicy,
 ) -> crate::Result<()> {
+    assert!(output_dir.exists());
     let reader = fs::File::open(&input_file_path)?;
     // Zip archives are special, because they require io::Seek, so it requires it's logic separated
     // from decoder chaining.
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -355,7 +355,16 @@ fn decompress_file(
     // Any other Zip decompression done can take up the whole RAM and freeze ouch.
     if formats.len() == 1 && *formats[0].compression_formats == [Zip] {
         let zip_archive = zip::ZipArchive::new(reader)?;
-        let _files = crate::archive::zip::unpack_archive(zip_archive, output_dir, question_policy)?;
+        let _files = if let ControlFlow::Continue(files) = smart_unpack(
+            Box::new(move |output_dir| crate::archive::zip::unpack_archive(zip_archive, output_dir, question_policy)),
+            output_dir,
+            &output_file_path,
+            question_policy,
+        )? {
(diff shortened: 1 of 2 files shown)


src/commands.rs
```diff
@@ -487,6 +516,57 @@ fn list_archive_contents(
...
+/// - If the archive contains only one file, it will be extracted to the `output_dir`
+/// - If the archive contains multiple files, it will be extracted to a subdirectory of the output_dir named after the archive (given by `output_file_path`)
+/// Note: This functions assumes that `output_dir` exists
+fn smart_unpack(
+    unpack_fn: Box<dyn FnOnce(&Path) -> crate::Result<Vec<PathBuf>>>,
+    output_dir: &Path,
+    output_file_path: &Path,
+    question_policy: QuestionPolicy,
+) -> crate::Result<ControlFlow<(), Vec<PathBuf>>> {
+    let temp_dir = tempfile::tempdir_in(output_dir)?;
```
marcospb19, line 529:
```suggestion
    assert!(output_dir.exists());
    let temp_dir = tempfile::tempdir_in(output_dir)?;
```

src/commands.rs
```diff
@@ -487,6 +516,57 @@ fn list_archive_contents(
...
+/// Unpacks an archive with some heuristics
+/// - If the archive contains only one file, it will be extracted to the `output_dir`
+/// - If the archive contains multiple files, it will be extracted to a subdirectory of the output_dir named after the archive (given by `output_file_path`)
+/// Note: This functions assumes that `output_dir` exists
+fn smart_unpack(
+    unpack_fn: Box<dyn FnOnce(&Path) -> crate::Result<Vec<PathBuf>>>,
+    output_dir: &Path,
+    output_file_path: &Path,
+    question_policy: QuestionPolicy,
+) -> crate::Result<ControlFlow<(), Vec<PathBuf>>> {
```
marcospb19, line 528:
```suggestion
) -> crate::Result<ControlFlow<(), Vec<PathBuf>>> {
    assert!(output_dir.exists());
```

sigmaSd:
fix #151

PR #210, opened 2021-11-16
Add progress to compressing/decompressing

sigmaSd:
This pr uses the input size as an approximation for the compressed archive size.

### DIFF
--- a/src/archive/tar.rs
+++ b/src/archive/tar.rs
@@ -15,15 +15,16 @@ use crate::{
     info,
     list::FileInArchive,
     utils::{self, Bytes},
-    QuestionPolicy,
 };
 
 /// Unpacks the archive given by `archive` into the folder given by `into`.
+/// Assumes that output_folder is empty
 pub fn unpack_archive(
     reader: Box<dyn Read>,
     output_folder: &Path,
-    question_policy: QuestionPolicy,
+    mut display_handle: impl Write,
 ) -> crate::Result<Vec<PathBuf>> {
+    assert!(output_folder.read_dir().expect("dir exists").count() == 0);
     let mut archive = tar::Archive::new(reader);
 
     let mut files_unpacked = vec![];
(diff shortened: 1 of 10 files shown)


sigmaSd:
Some thoughts:
- This appears to work well on my testing
- I think I should also add a flag (no-progress-bar) to disable this, this is not only for users but for possible future interactive tests (should be straight forward, since I prepared for that with the  display_handle api)
- Decompressing could also have this, but Its trickier because the output can be multiple files (so not sure what to monitor here)

Obviously this is subjective, Its up to you if its a good addition or not

src/archive/tar.rs
```diff
@@ -12,33 +12,29 @@ use walkdir::WalkDir;
...
     let mut files_unpacked = vec![];
     for file in archive.entries()? {
         let mut file = file?;
 
         let file_path = output_folder.join(file.path()?);
-        if !utils::clear_path(&file_path, question_policy)? {
-            // User doesn't want to overwrite
-            continue;
-        }
-
```
marcospb19, line 38: Are our overwrite questions still working?

marcospb19:
I'm worried this conflicts a lot with #197

src/archive/tar.rs
sigmaSd: unpack_archive is only calledby smart_which always create a new directory

So there is no files to overwrite

sigmaSd:
I can add a cli flag that restores the original behaviour, if this sound a good idea otherwise

marcospb19:
> I can add a cli flag that restores the original behaviour, if this sound a good idea otherwise

Yeah, that would make merging with --accessibility much easier.

SpyrosRoum:
Oh this is great, I just decompressed a relatively large thing and I thought it was stuck for a moment so I was about to see if I can implement something like this myself haha

sigmaSd:
That makes sense, I add `--disable-progress-bar` for now, so now the mechanism exists, the accessibility pr can just rename it, and repurpose it to its needs

sigmaSd:
Or you can merge the accessibility pr first and I'll rebase on top of it when I can

sigmaSd:
Btw does listing archive files needs a progress bar too?

sigmaSd:
Here is a re-based version https://github.com/ouch-org/ouch/compare/master...sigmaSd:acess?expand=1

Whats the point of `a11y_show` ? Shouldn't `log!` always check for `ACCESSIBLE`?

sigmaSd:
I went ahead and changed the flag to `ACCESSIBLE` in order to make the other pr conflicts as lest as possible

marcospb19:
> Whats the point of a11y_show ? Shouldn't log! always check for ACCESSIBLE?

So, `a11y_show` is meant to disable some `info!`, because `ouch` outputs too much text, and this is not accessible.

And the `ACCESSIBLE` global boolean is to change how the information is displayed.

I also got confused about it, but the former one hides text, and the latter reformats it.

marcospb19:
The `list` command probably does not need a progress bar.

I think I'll wait for the other PR to finish before merging this one, if it takes too long, I guess this one can be merged before, not sure if it really makes a difference, tho.

marcospb19:
Accessibility PR was merged :D, now this one is no longer blocked, if you wish to resolve the conflicts.

sigmaSd:
Done

Issue #183, opened 2021-11-10
Don't use absolute paths


marcospb19:
We can remove the `fs::canonicalize` function calls, as they are only being currently used to check if files exist, this type of check should be performed with `fs::metadata` instead.

This is currently causing the problem that our error messages have the absolute paths, some workaround functions could even be removed, like `strip_cur_dir` simplifying our code.

Example of absolute path in error messages:

![image](https://user-images.githubusercontent.com/38900226/141209941-efb080e5-283c-45ed-92d4-1e86884dd97c.png)

marcospb19:
I'll take a shot at fixing this.

SpyrosRoum:
Hey marcospb19, are you still working on this?

marcospb19:
Sadly not, I didn't even start it, I think this requires a medium refac just because it might break other parts of the code.

marcospb19:
I don't think this does make sense anymore, canonicalized is only called for files passed as CLI args, not unpacked files.

There seems to be no clear benefit to this, it would require reviewing a lot of code to check for errors, and it possibly introduces some bugs because the whole codebase expects input file paths to be absolute.

PR #214, opened 2021-12-07
Truncate long messages in the progress bar

sigmaSd:
Currently if the path is too long the progress bar wraps around, and starts spamming messages in new lines.

By using, `wide_msg`, long messages are automatically truncated by indicatf.

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -65,7 +65,7 @@ impl Progress {
                 let mut t = String::new();
                 t += "{wide_msg} [{elapsed_precise}] ";
                 if precise && current_position_fn.is_some() {
-                    t += "[{wide_bar:.cyan/blue}] ";
+                    t += "[{bar:.cyan/blue}] ";
                 } else {
                     t += "{spinner:.green} ";
                 }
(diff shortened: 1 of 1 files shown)


marcospb19:
There's something weird going on

image

image

sigmaSd:
Can you upload the file you're using because I can't reproduce this behavior.

sigmaSd:
Aside: I just noticed something, ouch fails to compress if a directory contains a broken symlink

sigmaSd:
Also what terminal are you using?

sigmaSd:
I can reproduce, I'll see if I can figure it out

sigmaSd:
I think its should be fixed now

marcospb19:
I'll run some more manual tests before merging.

sigmaSd:
I noticed that paths that are printed when extracting from zip or tar are canonicalized,  I think we need another pr to fix that which would improve the ux even more.

marcospb19:
You can create an issue or PR for the canonicalized paths, just have to check if that's not a duplicate of #183.

PR #215, opened 2021-12-07
support snappy format

### LINKED ISSUE #202: Support snappy format
https://google.github.io/snappy/

extension: `.sz`

crate: https://lib.rs/crates/snap

figsoda:
closes #202 

Also removed remaining mentions of incorrect lz extensions that wasn't removed in #198


marcospb19:
Nice, does the Snappy format has any MIME bytes to be add to the tests?

figsoda:
yes, application/x-snappy-framed, but infer doesn't support it
[file-format](https://lib.rs/file-format) supports it but is not nearly as popular, maybe we can try that
didn't check but [magic](https://lib.rs/magic) (libmagic binding) might support it as welll

marcospb19:
For reference, we have some MIME checks implemented here:
https://github.com/ouch-org/ouch/blob/bafc2d31b4dacf738ed6a82aa315df6ca2247fbb/src/utils/fs.rs#L65-L121

So, this could probably be reused in our tests?

If you wish to, I can merge this as it is, and then you can create an issue for covering the rest of it.

Or, you can try to solve everything in this PR.

figsoda:
I thought about reusing it by checking the warnings generated by the program, but #220 (just opened btw) might cause tests to fail

I don't plan on addressing this in this pull request

PR #217, opened 2021-12-07
Fix zip memory warnings

### LINKED ISSUE #178: Inconsistent `zip` compression and decompression memory usage warnings
image

The limitation of compressing zip archives with other streams only occurs when using multiple formats, like:

`archive.zip.gz`, or `archive.zip.xz.xz.xz.xz`.

`.zip` requires `io::Seek`, and it can only be provided in `fs::File` and `Vec<u8>`, so we can either compress directly to the final file, or load into a `Vec`, but not pipe to encoding streams.

Loading the whole file to a `Vec` might exhaust memory capacity, that's why we need to give those warnings in specific cases.

---

While filing this, I found out the `decompression` command has another warning with different formatting:

image

These messages formats should be consistent.

Crypto-Spartan:
resolves #178

also added functionality to ask user if they want to continue when working with `.zip` archives with extra compression formats (e.g. .zip.gz)

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -347,14 +347,10 @@ fn compress_files(files: Vec<PathBuf>, formats: Vec<Extension>, output_file: fs:
             writer.flush()?;
         }
         Zip => {
-            eprintln!("{yellow}Warning:{reset}", yellow = *colors::YELLOW, reset = *colors::RESET);
-            eprintln!("\tCompressing .zip entirely in memory.");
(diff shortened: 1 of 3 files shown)


marcospb19:
> also added functionality to ask user if they want to continue when working with .zip archives with extra compression formats (e.g. .zip.gz) 

That's an awesome addition, thank you so much for your contribution, we really appreciate it!

PR #219, opened 2021-12-07
Fix the hint suggestion for compressing multiple files

### LINKED ISSUE #218: hint suggestion for compressing multiple files has incorrect format
### Version

0.3.1

### Description

When executing the command `ouch compress 1 2 3 4 myarchive.zst` this is the output:

image

Notice that the last hint is incorrect in it's format.

### Current Behavior

_No response_

### Expected Behavior

The output should look like this instead:

image

### Additional Information

_No response_

Crypto-Spartan:
resolves #218

changed <string>.replace_range() to <string>.insert_str() since the "range" wasn't really a range. The start and end to the range was the same number.


marcospb19:
Thanks!

PR #221, opened 2021-12-08
Add redundant check for --yes and --no flags conflict

marcospb19:
Should be covered by clap, but we're double checking it now

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -33,12 +33,11 @@ impl Opts {
         | Subcommand::List { archives: files, .. }) = &mut opts.cmd;
         *files = canonicalize_files(files)?;
 
-        let skip_questions_positively = if opts.yes {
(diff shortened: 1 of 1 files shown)


src/cli.rs
```diff
@@ -33,12 +33,11 @@ impl Opts {
...
-            QuestionPolicy::AlwaysYes
-        } else if opts.no {
-            QuestionPolicy::AlwaysNo
-        } else {
-            QuestionPolicy::Ask
+        let skip_questions_positively = match (opts.yes, opts.no) {
+            (false, false) => QuestionPolicy::Ask,
+            (true, false) => QuestionPolicy::AlwaysYes,
+            (false, true) => QuestionPolicy::AlwaysNo,
+            (true, true) => unreachable!(),
```
marcospb19, line 40: Now we'll panic in case it fails.

PR #223, opened 2021-12-08
Add `--format` flag to specify formats

### LINKED ISSUE #134: Add `--format` flag
# Problem

Currently, we don't have control of what extensions should be used in compression or decompression, we rely on automatic extension detection.

Using `archive.tar.gz.xz` as an example, you might want to decompress it into:

    archive.tar.gz, without the .xz part, or
    archive.tar, without the .gz.xz part.

That is, decompress it partially, or just decompress a file without extensions.

# Workaround

Decompress and compress it again.

```
ouch decompress archive.tar.gz.xz -o result
cd result
ouch compress * .* archive.tar 
```

Note that the shell expansion .* is necessary in case you have any hidden files.

# New solution

`--format` flag that overwrites the automatic format detection system.

```sh
ouch compress file.tar file.tar.xz --format xz
ouch decompress secret_file --format tar.gz
```

We already have detection on partial compression:

```sh
ouch compress file.tar file.tar.xz
```

Ouch detects it should only partially compress the `.xz` part, and triggers it, I think this behavior is useful but can be triggered by accident.

We should leave this detection only as a warning to tell the users that if they want to partially compress, they should use the `--format` flag.

marcospb19:
Fixes #134.

- [x] `--format` flag to `compress` command.
- [ ] `--format` flag to `decompress` command.
- [ ] `--format` flag to `list` command.
- [ ] Do a ton of manual testing.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -10,7 +10,7 @@ use clap::Parser;
 use fs_err as fs;
 use once_cell::sync::OnceCell;
 
-use crate::{Opts, QuestionPolicy, Subcommand};
+use crate::{error::FinalError, Opts, QuestionPolicy, Subcommand};
 
 /// Whether to enable accessible output (removes info output and reduces other
 /// output, removes visual markers like '[' and ']').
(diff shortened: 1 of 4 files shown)


marcospb19:
I found out this is not the best moment to implement this feature, although it's highly wanted, I'm having some trouble with the current codebase, probably I should be doing some refactors before, maybe this will go to the 0.5 release instead, we'll see.

PR #224, opened 2021-12-08
Ignore broken symlinks when compressing

sigmaSd:


src/utils/fs.rs
```diff
@@ -119,3 +119,10 @@ pub fn try_infer_extension(path: &Path) -> Option<Extension> {
         None
     }
 }
+
+/// Returns true if a path is a symlink.
+/// This is the same as the nightly https://doc.rust-lang.org/std/path/struct.Path.html#method.is_symlink
+// Useful to detect broken symlinks when compressing. (So we can safely ignore them)
+pub fn is_symlink(path: &Path) -> bool {
+    fs::symlink_metadata(path).map(|m| m.file_type().is_symlink()).unwrap_or(false)
+}
```
marcospb19, line 128: I thought this was already stabilized, do we have to wait for stable 1.59 for accessing it?.

marcospb19 (approved):

marcospb19:
> as an alternative we can add a flag to preserve symlinks

Tar seems to support symlinks, but zip doesn't, so this flag would only work for tar?

figsoda:
`man zip` gave me this

```
--symlinks
       For  UNIX  and VMS (V8.3 and later), store symbolic links as such in the zip archive, instead of compressing and storing the file referred to by the link.
       This can avoid multiple copies of files being included in the archive as zip recurses the directory trees and accesses files directly and by links.
```

sigmaSd:
That doesn't mention broken symlinks explicitly and in practice if you try ziping a directory containing a broken symlink using `zip` it wont be in the zip archive

marcospb19:
I think I should then create an issue to add a `--symlink` flag that does the same, @sigmaSd do you also agree?

sigmaSd:
I guess its a simple flag and can be added easily

But I think it should be named -something like --dont--ignore--broken--symlinks 

My hunch is this is an advanced feature and the flag should be explicit like the above to show it

Or another idea is to wait until somone open an issue about missing broken symlinks and explain his usecase, then ir becomes obvious why we want this

marcospb19:
`--dont--ignore--broken--symlinks` is different from the `--symlink` I was thinking about.

The former is about throwing errors when symlinks are broken, and the latter is about not following symlinks when creating archives, adding the symlinks instead of the files they point to.

I think that the latter would be a more interesting and useful flag, could be named as `--store-symlinks`.

sigmaSd:
So I did some testing and there is actually some weird edge cases in how `zip` does this, but overall:
- zip follows symlinks by default (`-y` flag to not follow)
- tar doesn't follow symlinks by default(`-h` ? to follow)

Since ouch already is similar to zip in this area, I guess adding `--store-symlinks` is a good idea

marcospb19:
Thanks for your research!

src/utils/fs.rs
marcospb19, line 128: lol a couple months later this was stabilized.

https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html

Issue #212, opened 2021-11-23
List command waits for entire archive to be processed before outputting anything


AntonHermann:
Currently the list command reads the whole archive into a `Vec` before outputting anything, thus for big archives a long time nothing is printed at all

## Expected behavior
- when running without `--tree`: print each file as soon as it's processed
- when running with `--tree`: show a progress indication ("reading files..." or "reading files: _XX_%") while files are read

## Implementation suggestions
- when running without `--tree`: instead of returning a `Vec` from _archive::{tar,zip}::list_archive_, maybe we can return an iterator or similar instead?
- when running with `-tree`: gray text which is deleted as soon as output is ready would look nice I guess

marcospb19:
Like the idea.

But what happens with text that is deleted afterwards when the output is redirected to another file? Does it break anything? If so, I think that a `[INFO]` call would be better.

AntonHermann:
Good question, I'll try to figure out how that is handled and how other CLI programs that rewrite their output handle this case

PR #225, opened 2021-12-08
List command: print file immediatly after it is processed

sigmaSd:
fix https://github.com/ouch-org/ouch/issues/212

If performance is a concern , I can make the listing happen in another thread, and each file processed is sent to a shared queue, the printing just pops from the queue each time something is available. This makes the listing not wait for printing to Stdout. But I thought it may not matter that much.


sigmaSd:
Regarding the last commit, I glanced the code of tar-rs, and adding `into_entries` doesn't look feasible without major rewrite. I think leaking is not that bad though.

Crypto-Spartan:
After reading the documentation for [Box::leak()](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.leak), is this really the best way to do this? Per the docs:

"This function is mainly useful for data that lives for the remainder of the program’s life. Dropping the returned reference will cause a memory leak. If this is not acceptable, the reference should first be wrapped with the Box::from_raw function producing a Box. This Box can then be dropped which will properly destroy T and release the allocated memory."

At the very least, wouldn't it be best to safely drop the leaked box by first wrapping it in Box::from_raw? 

I feel like Box::leak should be avoided here if possibe as leaking memory goes against the Rust paradigm...

sigmaSd:
The last commit uses a new thread for listing , this gives it best performance since it wont wait for println io , and removes the leak as well.

The only thing I'm not sure about is the "info!(Processing: " line, I think its too verbose, Maybe its better if I made it print on the same line there (with '\r")?

marcospb19:
> The only thing I'm not sure about is the "info!(Processing: " line, I think its too verbose, > Maybe its better if I made it print on the same line there (with '\r")?

Yeah that is too verbose, I think you can just remove it.

If we are going to print something there, probably should just be the name of the archive that will be shown.

sigmaSd:
I changed it to use indicatif, I think it looks good like this, this is similar to what was originally suggested in the issue.

marcospb19:
Looking great!!!!!

Issue #114, opened 2021-10-19
Redirect all text output to stderr and add flag to suppress non-errors


marcospb19:
_Originally posted by @marcospb19 in https://github.com/ouch-org/ouch/issues/110#issuecomment-946958436_

# Steps

1. Use `eprint` instead of `print`.
2. add `-q/--quiet` flag.

# Reasoning:

1. Will make it possible for us to use STDOUT as the output of compression and decompression.
2. With everything in the `stderr`, the user loses the ability of redirecting only the errors to a log file, by adding the `--quiet` flag, we provide a decent way of logging it.

---

Example of redirection:

```sh
ouch decompress -q file.xz 2> ouch.log
```

marcospb19:
From https://github.com/ouch-org/ouch/issues/110#issuecomment-946961949:

in addition to a quiet flag, we can also use [tracing](https://lib.rs/crates/tracing) with [EnvFilter](https://docs.rs/tracing-subscriber/*/tracing_subscriber/struct.EnvFilter.html) from [tracing-subscriber](https://lib.rs/crates/tracing-subscriber)

marcospb19:
I would love to see `tracing` being used in `ouch` for this, even tho I have no prior experience with it.

marcospb19:
The first point kills it, no `tracing` then :thinking:.

SpyrosRoum:
Here is another option. 
Two simple declarative macros:
```rust
macro_rules! stdout_println {
    ($($arg : tt) *) => {
        if atty::is(atty::Stream::Stdout) {
            println!($($arg)*);
        }
    };
}

macro_rules! stdout_print {
    ($($arg : tt) *) => {
        if atty::is(atty::Stream::Stdout) {
            print!($($arg)*);
        }
    };
}
```

So we can use the usual `println` for when we want to definetly output something, `eprintln` for errors, and the macros for logs and other info.

Thoughts?

figsoda:
I still prefer only compressing and decompressing to stdout and printing everything else to stderr

figsoda:
Could you elaborate on what could be confusing/annoying about printing all the logs and messages to stderr?

SpyrosRoum:
.. This is what I'm trying to say.

Running the first command produces this if successful:
`[INFO] Successfully compressed 'foo.xz'.`
and maybe it will produce this too
`Do you want to overwrite 'foo.xz'? [Y/n]`

If we print all that to stderr then using `2>` will not show either of them,

[...]

Issue #110, opened 2021-10-19
Support decompressing to `stdout`


marklit:
Normally passing a dash when decompressing with other tools will cause the decompressor to output to stdout. There doesn't seem to be such support in ouch.

```bash
echo hello > hello.txt
ouch c hello.txt hello.tar.lzma.zst
ouch d hello.tar.lzma.zst - | hexdump -C
```

```
00000000  44 6f 20 79 6f 75 20 77  61 6e 74 20 74 6f 20 6f  |Do you want to o|
00000010  76 65 72 77 72 69 74 65  20 27 2e 2f 68 65 6c 6c  |verwrite './hell|
00000020  6f 2e 74 78 74 27 3f 20  5b 1b 5b 33 38 3b 35 3b  |o.txt'? [.[38;5;|
00000030  31 30 6d 59 1b 5b 33 39  6d 2f 1b 5b 33 38 3b 35  |10mY.[39m/.[38;5|
```

I also tried just compressing a single file as a ZSTD stream instead of tar'ing the file first and ouch crashed. #96 probably covers this issue.

```bash
ouch c hello.txt hello.zst
```

```
thread 'main' panicked at 'internal error: entered unreachab`le code', src/commands.rs:165:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

```bash
ouch c hello.txt hello.lzma
```

```
thread 'main' panicked at 'internal error: entered unreachable code', src/commands.rs:165:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

SpyrosRoum:
I think decompressing to stdout would be a good feature, similar to #6.
About the crash there is a pr open that fixes it.

figsoda:
we should also move all current outputs to stderr, there are still a couple uses of `println` in the code base

marcospb19:
@figsoda great point, everything would need to go to `stderr`, otherwise, decompressing to `stdout` will corrupt all the files.

figsoda:
`ouch d file.xz 2>ouch.log` should work if they are outputted to stderr

marcospb19:
Moving this specific part of the discussion to the dedicated issue #114.

PR #226, opened 2021-12-08
simple eprintln fixes

Crypto-Spartan:
I found 2 instances where ouch should be using eprintln instead of println and fixed them here.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -55,12 +55,12 @@ pub fn _info_helper(handle: &mut impl std::io::Write) {
     write!(handle, "{}[INFO]{} ", *YELLOW, *RESET).unwrap();
 }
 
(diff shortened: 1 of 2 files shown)


marcospb19:
Have you reviewed all the usages of `println` and `eprintln`?

We had discussions previously about where output should be directed to, and then stuff changed, I think that these eprintln were left from previous decisions.

Crypto-Spartan:
I saw the issues with the discussion, notably #110 & #114. 

The discussions in those issues talked about moving all output to stderr, even normal `[INFO]` outputs. In the 2 instances I changed, they should have been directed to stderr in the first place (since they were actual errors) but they were going to stdout instead. So this PR shouldn't have any impact on the other issues, AFAIK.

PR #227, opened 2021-12-08
remove redundant user_wants_to_continue function

Crypto-Spartan:
after #217, there were 2 functions in utils/questions.rs that were nearly identical, the only difference being the string passed to the user for confirmation. This PR merges the functions into a single function and creates an enum (`QuestionAction`) to pass to the function to choose the respective compression/decompression option for the string.


Crypto-Spartan:
The check that failed was with `cargo fmt`. I added another commit to make it happy.

PR #230, opened 2021-12-14
Mark directories when compressing to zip regardless of their contents

sigmaSd:
To test this:
- compress a non empty directory with `zip`
- run: `ouch l dir.zip` -> it outputs the dir name correctly (in blue)
- compress a non empty directory with `ouch`
- run: `ouch l dir.zip` -> it doesn't output the dir name, because it doesn't identify it as a directory

### DIFF
--- a/src/archive/zip.rs
+++ b/src/archive/zip.rs
@@ -14,10 +14,7 @@ use crate::{
     error::FinalError,
     info,
     list::FileInArchive,
-    utils::{
-        self, cd_into_same_dir_as, concatenate_os_str_list, dir_is_empty, get_invalid_utf8_paths, strip_cur_dir,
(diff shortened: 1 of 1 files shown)


marcospb19:
I'm not being able to reproduce this error, but I'm trying only with a single empty folder, how can I see it?

The output here is the same for ouch and zip.

I tried ouch in the latest master branch, as well as the PR branch.

sigmaSd:
The folder has to be **not** empty

mkdir a; touch a/b;
ouch c a ouch_a.zip; zip -r a.zip a

ouch l a.zip and ouch l a_ouch.zip should show differently

marcospb19:
Right, sorry for that one, noticed the fix now.

Merging!

PR #236, opened 2021-12-30
Make yes, no, and accessible flags global

### LINKED ISSUE #233: Accept flags in any position
Currently, if we type:
```shell
ouch decompress archive.zip --yes
```

`ouch` will complain that this flag does not exist, instead, you should write it like this:

```shell
ouch --yes decompress archive.zip
```

Would be nicer if the the flags `--yes`, `--no`, and `--accessible`, plus the incoming `--store-symlinks`, `--gitignore` and `--format`, could be added in any place.

LovecraftianHorror:
Closes #233

`clap` allows for specifying that a flag is global meaning that it propagates down to all child subcommands as well. This makes it so that the following are equivalent `ouch compress --yes ...` or `ouch --yes compress ...`

### DIFF
--- a/src/opts.rs
+++ b/src/opts.rs
@@ -12,15 +12,15 @@ use clap::{Parser, ValueHint};
 #[clap(version)]
 pub struct Opts {
     /// Skip [Y/n] questions positively.
-    #[clap(short, long, conflicts_with = "no")]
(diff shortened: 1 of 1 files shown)


marcospb19:
That's great!

Thanks for your contribution and sorry for the delay, we have been busy lately.

PR #238, opened 2022-01-01
Cleanup integration tests

LovecraftianHorror:
There were a few different changes here:

- 5775d4f26804433d81dc86f61ac22df2e04992ec made it so that the dummy files are actually filled with contents. `.fill_bytes()` will overwrite existing contents. It won't push on any new data even if there is capacity for it
- 1ad3652a40c17c4945863eda35f22a9799c436ca just switched to some nicer syntax for generating a random value in a range
- c5e9696f8ddfda02eb57ac9c6c2e6fb77d0b4e34 added extensions to test for all the formats listed in the README
- f3d564f8b00ccb9f407f6130240e772449efacd3 runs the integration tests with the accessibility flag to drastically reduce the runtime
  - This change brought the time to run integration tests down from around 2 minutes to less than 30 seconds for me
  - I went through a lot of effort trying to track down what specifically was slow here. My current best guess is that there is that `indicatif` runs much slower in a spawned command for some reason. I'll be spending some more time trying to track down the root cause of this


vrmiguel:
Awesome :D 

Thanks!

PR #239, opened 2022-01-04
fix warning for zip with additional formats

### LINKED ISSUE #231: Warning for compressing with Zip with additional formats is thrown incorrectly
### Version

35a158ec63aa88004e4daed078aa978aa1ff95ab

### Description

image

### Current Behavior

Shows this warning.

### Expected Behavior

Should not show this warning.

### Additional Information

This warning should only be shown on formats composed with multiple formats after zip:

```
a.zip.gz
a.zip.xz
a.zip.gz.xz
```

Crypto-Spartan:
resolves #231

Additionally, I changed the return type of compress_files() from Result<()> to Result\<bool\>.

Previously, if a user had selected no when prompted to continue compression: 
* there would still be an output message that said compression was successful
* there would be a leftover archive file with nothing in it

This is an example of what the message used to look like:
image

The fix removes the incorrect `INFO` message when a user selects `n` as well as deletes the empty archive file.


marcospb19:
Did minor changes and had to solve conflicts twice :V

Waiting for CI to merge.

PR #241, opened 2022-01-05
remove lib.rs and move contents to main.rs

### LINKED ISSUE #234: Remove `lib.rs` and move stuff to `main.rs`
`ouch` currently has an `lib.rs` and `main.rs` file, which increases the compile time.

First it compiles the `lib.rs` crate part and then it compiles `main.rs`, plus, both operations cannot be parallelized.

By only having `main.rs` I think that the compilation time will improve a bit.

Crypto-Spartan:
resolves #234


vrmiguel:
Thanks!

PR #245, opened 2022-01-16
feat: allow ignoring hidden files and files matched by .gitignore files

vrmiguel:
This PR replaces our usage of `walkdir` by `ignore`, which allows us to ignore hidden files and files matched by `.gitignore` files.

`ignore` also has the ability of creating parallel iterators, which might prove useful in the future


marcospb19 (approved):
Awesome!

PR #249, opened 2022-02-03
feat: don't allocate when possible in `to_utf`, `nice_directory_display`

vrmiguel:
Paths are overwhelmingly more likely to be valid UTF-8 rather than not, so using `Cow` here is likely very beneficial


marcospb19 (approved):
awesome! (run clippy and merge :D)

PR #250, opened 2022-02-06
feat: recover last modified time when unpacking zip archives

vrmiguel:


src/archive/zip.rs
```diff
@@ -33,7 +33,9 @@ where
     D: Write,
 {
     assert!(output_folder.read_dir().expect("dir exists").count() == 0);
-    let mut unpacked_files = vec![];
+
+    let mut unpacked_files = Vec::with_capacity(archive.len());
```
marcospb19, line 37: gotta go fast

marcospb19 (approved):

PR #251, opened 2022-02-13
chore: allow overriding the completions output directory

jcgruenhage:
This makes it a bit easier for people packaging ouch to get to the completions, as they don't end up hidden away in `target/${RUST_TARGET}/release/build/ouch-<some-hash>/out/completions`. It'd also be possible to maintain patches like these downstream in the distro packaging instead, but I feel like this change makes sense to include here.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,4 +1,4 @@
-use std::{env, fs::create_dir_all, path::Path};
+use std::{env, fs::create_dir_all};
 
 use clap::{ArgEnum, IntoApp};
 use clap_complete::{generate_to, Shell};
@@ -12,7 +12,11 @@ fn main() {
         return;
     }
 
-    let out = &Path::new(&env::var_os("OUT_DIR").unwrap()).join("completions");
+    let out = &env::var_os("OUCH_COMPLETIONS_DIR")
+        .map(|path| PathBuf::from(&path))
+        .or_else(|| env::var_os("OUT_DIR").map(|path| PathBuf::from(&path)).map(|path| path.join("completions")))
+        .unwrap();
+
     create_dir_all(out).unwrap();
     let app = &mut Opts::into_app();
 
--- a/build.rs
+++ b/build.rs
@@ -1,4 +1,39 @@
-use std::{env, fs::create_dir_all};
+/// This build script checks for env vars to build ouch with shell completions.
+///
+/// # How to generate shell completions:
+///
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -1,3 +1,19 @@
...
+/// Building ouch and completions (need to be set to `'1'`):
+/// ```sh
+/// GEN_COMPLETIONS=1 cargo build
+/// ```
+///
+/// Building ouch and completions, inside another folder:
+///
+/// ```sh
+/// GEN_COMPLETIONS=1 OUCH_COMPLETIONS_DIR=ouch-completions cargo build
+/// ```
```
marcospb19, line 14: While doing this, one question came to mind, what if instead of two env vars, I just check for `GEN_COMPLETIONS`, if it's value is `1` or `true`, the code uses `OUT_DIR`, if it's another value, we use as the destination folder.

So usage would be:

```sh
# simple build completions 
GEN_COMPLETIONS=1 cargo build
```

```sh
# to new folder 
GEN_COMPLETIONS=ouch-completions cargo build
```

Waiting for your feedback on this, thanks in advance!

marcospb19 (commented):
You inspired me to try to document this better to help people package `ouch`.

build.rs
jcgruenhage: Sounds good to me, yeah. Probably the most elegant way to handle this.

build.rs
```diff
@@ -1,3 +1,38 @@
+/// This build script checks for env vars to build ouch with shell completions.
+///
+/// # How to generate shell completions:
+///
+/// Set `OUCH_COMPLETIONS_FOLDER` to the name of the destination folder:
+///
+/// ```sh
+/// OUCH_COMPLETIONS_FOLDER=my-folder cargo build
+/// ```
```
marcospb19, line 9: @jcgruenhage sorry for the delay on this, I'm trying to get back to ouch.

The final code ended up being a bit different from what we agreed, I maintained retrocompability to keep older packages of ouch working.

Everything is now documented in this file.

marcospb19: @jcgruenhage will you need to update something in the new void package?

jcgruenhage: Looks good to me. And yeah, will need adjustment once a new release of ouch is out. Both of the patches can be dropped and the pre_build phase will need to be adjusted to the new way these env vars should be working, but that shouldn't be too much work.

cc: @ElDifinitivo

PR #252, opened 2022-02-13
fix: remove single quotes from clap doc comments

jcgruenhage:
Apparently, clap_complete doesn't properly escape single quotes in doc
comments, leading to broken shell completions for zsh. To circumvent
this, I've removed the single quote in the one place where it shouldn't
have been in the first place ("it's" vs "its", contraction vs possessive
pronoun), and replaced "it's" with "it is" in the other place.

Reference to upstream issue: https://github.com/clap-rs/clap/issues/1596


marcospb19 (approved):
Thanks!

PR #253, opened 2022-02-28
Update dependencies

Crypto-Spartan:
I took a look at Ouch's dependencies and realized that there were a bunch that could use an update. Most of the dependency updates didn't require any changes elsewhere, except for clapv3.1.3. I made a small change in the build.rs script as we were using a function that clap has recently deprecated.


Crypto-Spartan:
each commit is very small, if it looks like a lot of changes it's just due to changes of the Cargo.lock file. I made each update a separate commit incase an update happened to break something

marcospb19:
Thanks a lot!

Issue #268, opened 2022-08-17
Improve error message description when trying to compress a folder with a multiple single file formats


marcospb19:
### Version

e377dd3 (latest)

### Description

Ouch is not giving a correct error message when trying to compress a folder with multiple single formats:

```
$ ouch compress linux.pkg linux.xz.gz.zst
```

### Current Behavior

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/commands/mod.rs:102:49
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Expected Behavior

Something like this, when we just pass ".xz".

```
$ ouch compress linux-5.17.5.arch1-1-x86_64.pkg linux.xz
[ERROR] Cannot compress to 'linux.xz'.
 - You are trying to compress multiple files.
 - The compression format 'xz' cannot receive multiple files.
 - The only supported formats that archive files into an archive are .tar and .zip.

hint: Try inserting '.tar' or '.zip' before 'xz'.
hint: From: linux.xz
hint: To:   linux.tar.xz
```

### Additional Information

It would be beneficial, while solving this, to write integration tests that check what is the output given.

marcospb19:
This is still not working properly:

```
ouch c empty-folder/ a.gz
[ERROR] failed to read from file `/home/marcospb19/ouch/empty-folder`
```

I think I have detected where the problem is tho.

PR #269, opened 2022-08-18
Fix wrong filename suggestions when compressing folders into non-archives

vrmiguel:
Hopefully solves #268 but doesn't add the integration test since I'm not sure how that should be done

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -16,7 +16,7 @@ use utils::colors;
 use crate::{
     commands::{compress::compress_files, decompress::decompress_file, list::list_archive_contents},
     error::FinalError,
-    extension::{self, flatten_compression_formats, Extension},
+    extension::{self, flatten_compression_formats, Extension, SUPPORTED_EXTENSIONS},
     info,
     list::ListOptions,
     utils::{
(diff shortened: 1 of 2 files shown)


src/commands/mod.rs
```diff
@@ -46,6 +46,40 @@ fn represents_several_files(files: &[PathBuf]) -> bool {
     files.iter().any(is_non_empty_dir) || files.len() > 1
 }
 
+/// Builds a suggested output file in scenarios where the user tried to compress
+/// a folder into a non-archive compression format, for error message purposes
+///
+/// E.g.: `build_suggestion("file.bz.xz", ".tar")` results in `Some("file.tar.bz.xz")`
+fn build_suggestion(path: &Path, suggested_extension: &str) -> Option<String> {
```
marcospb19, line 53: Could you change the name of this function to be more specific? Ouch currently does various suggestions inside of this mod.rs, this one is just one of them.

marcospb19 (changes requested):

src/commands/mod.rs
vrmiguel: Done

marcospb19 (approved):
:crab: :heart:

PR #271, opened 2022-09-01
fix: respect file permissions when compressing zip files

figsoda:
Only tested on linux

on unix platforms: keeps file permissions

on non-unix platforms: sets unix permissions to 755 if the file is executable (I'm not sure if I should keep this because iirc ouch doesn't respect permissions when decompressing zip files on non-unix platforms)


marcospb19:
So, I have been thinking about dropping non-unix (and Windows) support just because I, @vrmiguel, and most other developers cannot test and ensure everything is working there.

I also think that all Ouch's users are using unix anyways.

What do you think @vrmiguel?

figsoda:
any updates on this?

marcospb19:
Thanks for the ping, I spoke to Vrmiguel and we both agreed on dropping Windows support.

I think this PR can be merged as is and then I can create a new issue for dropping the support as a whole, what do you think?

figsoda:
I would be a little hesitant on completely dropping Window support, but I think we could official drop support for Windows and still make it compile on Windows.

We can merge this and get more opinions on this in a new issue.

PR #273, opened 2022-10-03
Generate man pages with clap_mangen

figsoda:
* Update dependencies (most notably clap 4)
* Generate man pages using clap_mangen (updated ci accordingly as well)
* Apply clippy lints
* I realized I put some of my changelog entries in the wrong version in previous pull requests so I moved them to the correct places

This breaks backwards compatibility for maintainers (not users)

I didn't update indicatif in this pr because breaking changes were introduced, I will do that in a separate pr


marcospb19 (approved):
Sorry for the delay, it is looking neat!

image

The weird thing with this man page generation is that flags like `--yes` do not expect a value.

```
$ cargo run -q -- --yes=true
error: The value 'true' was provided to '--yes' but it wasn't expecting any more values

Usage: ouch --yes <COMMAND>

For more information try '--help'
```

We'll see if it is fixed in next versions of clap.

PR #274, opened 2022-10-11
Check for errors when writing file times

marcospb19:
This PR adds error treatment for when access and modification
times are written when decompressing an archive.


figsoda (approved):
changes lgtm
another options is to use [filetime](https://lib.rs/crates/filetime) which tar is also using, so we have less `unsafe`s to maintain

marcospb19:
@figsoda I force-pushed a new implementation that uses `filetime` and removes the two `unsafe` uses in this file.

figsoda:
sorry, muscle memories

PR #277, opened 2022-10-11
Warn user if file extension is passed as file name

marcospb19:
example: `.tar.gz` is a file named .tar with extension `.gz`

this can be confusing because people might expect a
path `'.tar.gz'` to be a `.tar.gz` archive, but it's not

### DIFF
--- a/src/extension.rs
+++ b/src/extension.rs
@@ -3,6 +3,7 @@
 use std::{ffi::OsStr, fmt, path::Path};
 
 use self::CompressionFormat::*;
+use crate::warning;
 
 /// A wrapper around `CompressionFormat` that allows combinations like `tgz`
 #[derive(Debug, Clone, Eq)]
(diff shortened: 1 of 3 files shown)


marcospb19:
Branch name was supposed to be `waRn-file-extension-passed-as-file-name`, not `Wann`, I think if I change it now the PR will break so I'll leave it like this xD

src/extension.rs
```diff
@@ -103,19 +104,25 @@ pub const SUPPORTED_EXTENSIONS: &[&str] = &[
...
-    // if path.file_name().is_some() && all.iter().any(|ext| path.file_name().unwrap() == *ext) {
-    //     todo!("we found a extension in the path name instead, what to do with this???");
-    // }
-
     let mut extensions = vec![];
 
+    if let Some(file_stem) = path.file_stem().map(Path::new).and_then(Path::to_str) {
+        let file_stem = file_stem.trim_matches('.');
+
+        let valid_extensions = [
```
figsoda, line 116: maybe we can use `SUPPORTED_EXTENSIONS` here?

src/extension.rs
```diff
@@ -103,19 +104,25 @@ pub const SUPPORTED_EXTENSIONS: &[&str] = &[
...
-    // // TODO2: warn the user that currently .tar.gz is a .gz file named .tar
-    //
-    // let all = ["tar", "zip", "bz", "bz2", "gz", "xz", "lzma", "lz"];
-    // if path.file_name().is_some() && all.iter().any(|ext| path.file_name().unwrap() == *ext) {
-    //     todo!("we found a extension in the path name instead, what to do with this???");
-    // }
-
     let mut extensions = vec![];
 
+    if let Some(file_stem) = path.file_stem().map(Path::new).and_then(Path::to_str) {
```
figsoda, line 113:
```suggestion
    if let Some(file_stem) = path.file_stem().and_then(OsStr::to_str) {
```

figsoda:
Unrelated to this, but using [split](https://doc.rust-lang.org/std/primitive.slice.html#method.split) might have some performance gain over what we have right now, especially when we have to do `extensions.reverse()`

marcospb19:
> Unrelated to this, but using [split](https://doc.rust-lang.org/std/primitive.slice.html#method.split) might have some performance gain over what we have right now, especially when we have to do extensions.reverse()

Yeah, but in the context of compressing and decompressing files, I think that the performance gain from parsing the path is negligible (this also runs only for CLI given args I think).

figsoda (approved):

PR #279, opened 2022-10-12
set last modified time during zip compression

figsoda:
followup to #278


marcospb19 (approved):

marcospb19:
lgtm, we don't have automated tests for this so if you have manually tested if it's working I think you can merge

PR #280, opened 2022-10-12
rewrite progress module

figsoda:
- removes all remaining uses of unsafe
- no more mpsc channels
- no more display handle abstractions
- less passing around functions

this should behave the same as before, but I haven't extensively tested it, mostly been relying on the integration tests

### DIFF
--- a/src/archive/zip.rs
+++ b/src/archive/zip.rs
@@ -27,11 +27,7 @@ use crate::{
 
 /// Unpacks the archive given by `archive` into the folder given by `output_folder`.
 /// Assumes that output_folder is empty
-pub fn unpack_archive<R, D>(
-    mut archive: ZipArchive<R>,
-    output_folder: &Path,
-    mut display_handle: D,
-) -> crate::Result<Vec<PathBuf>>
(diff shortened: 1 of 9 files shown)


src/archive/zip.rs
```diff
@@ -27,11 +27,7 @@ use crate::{
 
 /// Unpacks the archive given by `archive` into the folder given by `output_folder`.
 /// Assumes that output_folder is empty
-pub fn unpack_archive<R, D>(
-    mut archive: ZipArchive<R>,
-    output_folder: &Path,
-    mut display_handle: D,
-) -> crate::Result<Vec<PathBuf>>
+pub fn unpack_archive<R, D>(mut archive: ZipArchive<R>, output_folder: &Path, mut out: D) -> crate::Result<Vec<PathBuf>>
```
marcospb19, line 30: I find the name of the parameter `out` to be a little misleading just because one could assume it's the `out`_put_ of the function.

While it's actually our stdout control "thing", where we log stuff to the user.

Could you give this a less ambiguous name? Example: `log_out`.

marcospb19 (changes requested):
This PR is a huge improvement in code quality, it's much much simpler now.

Code is amazing, just one nitpick:

marcospb19 (approved):

PR #295, opened 2022-10-16
Fix not overwriting a folder when compressing

marcospb19:
And also rename `ask_to_create_file` function.


marcospb19:
Very small refac, the functions in the question.rs suffer from bad naming because most of them do more than one thing, so it's weird to give them a proper name.

figsoda (approved):
lgtm

marcospb19:
This also solves a bug, so I will rename the PR and add it to the CHANGELOG as a bug fix.

figsoda:
btw, now ci takes a long time for the cross compilations to complete, so you might want to skip ci in the changelog update if you want to save some time

PR #307, opened 2022-11-18
update README for 0.4.0 release

marcospb19:


README.md
```diff
@@ -58,23 +60,56 @@ ouch decompress summer_vacation.zip --dir pictures
...
 ```sh
-# Compress four files
-ouch compress 1 2 3 4 archive.zip
+# Compress two files into `archive.zip`
+ouch compress one.txt two.txt archive.zip
 
-# Short alternative
-ouch c file.txt file.zip
+# Compress text.txt using .lz4 and .zst
+ouch compress text.txt text.txt.lz4.zst
```
figsoda, line 70: this is a little more readable to me
```suggestion
# Compress file.txt using .lz4 and .zst
ouch compress file.txt file.txt.lz4.zst
```

figsoda (commented):
benchmarks should probably be updated to 0.4.0, also it would be nice to show the command ran and the revision of linux

marcospb19:
> benchmarks should probably be updated to 0.4.0, also it would be nice to show the command ran and the revision of linux

I used `hyperfine` for the benchmark but I don't recall exactly what was the command, will be opening an issue to remember to update that.

That said, I think it is not good to update the benchmarks for each release, so probably we can also remove the version at which the benchmark was done.

README.md
marcospb19: Did it.

PR #310, opened 2022-11-23
remove installation script

marcospb19:
The installation script was cumbersome to maintain and was
fishy because it required root permissions to install Ouch.

Now that our releases contain artifacts for each target, it
becomes easier for users to download Ouch and install it.


README.md
```diff
@@ -132,16 +132,9 @@ yay -S ouch
 cargo install ouch
 ```
 
-## Using our script to download the latest binary
+## Download the latest release tarball
```
figsoda, line 135: it's `zip` files for windows, so maybe `archive` would be more accurate?

figsoda (approved):

PR #311, opened 2022-11-23
Check for EOF when asking questions

marcospb19:
When invoking Ouch from shell scripts, it is likely that questions will
be left unanswered with EOF, this PR adds a check to prevent Ouch
from interpreting EOF as Yes in Y/N questions.

image


figsoda:
looks like theres an issue with tests, might need to add `-y` the ouch calls

figsoda (approved):

Issue #296, opened 2022-10-16
Allow Ouch to decompress an archive into an existing folder


marcospb19:
Running

```
ouch decompress archive.tar --dir out
```

Asks the user if they want to delete the previous contents of `out` and overwrite them with the new contents from the archive.

Maybe it should add the new contents to the already existing folder, asking for overwrites every time some file conflicts (this is already the behavior when not using the `-d/--dir` flag)

a-moreira:
hey @marcospb19, I'd like to do this as a starter  :-)
I've created a PR to do this, but it's rather straightforward I guess. Let me know if something is wrong or if you want any other stuff to be implemented: https://github.com/ouch-org/ouch/pull/321

PR #321, opened 2022-12-10
allow ouch to decompress archive into existing folder

a-moreira:
closes https://github.com/ouch-org/ouch/issues/296

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -236,10 +236,6 @@ pub fn run(
             // The directory that will contain the output files
             // We default to the current directory if the user didn't specify an output directory with --dir
             let output_dir = if let Some(dir) = output_dir {
-                if !utils::clear_path(&dir, question_policy)? {
-                    // User doesn't want to overwrite
-                    return Ok(());
-                }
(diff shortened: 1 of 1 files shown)


figsoda (approved):
This doesn't ask recursively for files in subdirectories, but this also seems to the default behavior as mentioned in the original comment

marcospb19:
> This doesn't ask recursively for files in subdirectories, but this also seems to the default behavior as mentioned in the original comment

From my tests it seems that it is asking recursively, however, there are some surprises when using `--dir` and triggering the `smart unpack` functionality.

With `--dir` the output is redirected to another folder, but if you're unpacking an archive with more than one file inside, it will unpack it inside of an extra nested folder.

I think you thought it was not asking to overwrite because it was unpacked on another folder, so no paths conflicted.

This is probably talk for another issue or PR, _smart unpack_ should probably be disabled when `--dir` is present (or even completely removed from Ouch).

---

Thanks @a-moreira for your PR! :tada: :tada: :tada:

marcospb19:
@a-moreira I forgot to tell you to add an entry to CHANGELOG.md, I'll be making a commit with your name in there.

Issue #294, opened 2022-10-15
Add flag `-q/--quiet`


marcospb19:
`-q` flag description: "Silent all non-error messages."

EDIT: maybe `-q/--quiet` should make output less verbose, but keep the most useful stuff.

a-moreira:
hey @marcospb19 , the idea here would be to use the `clap_verbosity_flag` [crate](https://docs.rs/clap-verbosity-flag/latest/clap_verbosity_flag/index.html), or to modify the macros etc?

marcospb19:
@a-moreira we don't need `clap_verbosity_flag` here because by adding `--quiet` we'll have just two levels of verbosity, and that can be represented with just a boolean.

I was thinking about your suggestion, because we could add a `-v/--verbose` flag and use this crate, but I don't think it would be a good idea, every time I look up a `tar` usage example, people add the `-v` to make the output verbose and show all files that are being decompressed, so I think it's a good default as it is (at the time of writing, Ouch does this by default).

marcospb19:
I created this issue in tens of seconds and didn't give it much thought, now that I think about it, it's better to have `-q/--quiet` suppress only the verbose output, like the list of files that are packed and unpacked from and into archives, but we should keep outputting the rest.

Without `--quiet`:
```
$ ouch c src archive.tar.gz
[INFO] Compressing 'src'.
[INFO] Compressing 'src/archive'.
[INFO] Compressing 'src/archive/mod.rs'.
[INFO] Compressing 'src/archive/tar.rs'.
[INFO] Compressing 'src/archive/zip.rs'.
[INFO] Compressing 'src/utils'.
[INFO] Compressing 'src/utils/colors.rs'.
[INFO] Compressing 'src/utils/file_visibility.rs'.
[INFO] Compressing 'src/utils/formatting.rs'.
[INFO] Compressing 'src/utils/mod.rs'.
[INFO] Compressing 'src/utils/fs.rs'.
[INFO] Compressing 'src/utils/question.rs'.
[INFO] Compressing 'src/commands'.
[INFO] Compressing 'src/commands/list.rs'.
[INFO] Compressing 'src/commands/compress.rs'.
[INFO] Compressing 'src/commands/decompress.rs'.
[INFO] Compressing 'src/commands/mod.rs'.
[INFO] Compressing 'src/accessible.rs'.
[INFO] Compressing 'src/error.rs'.
[INFO] Compressing 'src/list.rs'.
[...]

PR #325, opened 2022-12-19
add cli option to (de)compress quietly

a-moreira:
closes https://github.com/ouch-org/ouch/issues/294

### DIFF
--- a/src/opts.rs
+++ b/src/opts.rs
@@ -30,7 +30,7 @@ pub struct Opts {
     pub hidden: bool,
 
     /// Silences output
-    #[arg(short = 'Q', long)]
+    #[arg(short = 'q', long)]
     pub quiet: bool,
 
     /// Ignores files matched by git's ignore files
--- a/src/opts.rs
+++ b/src/opts.rs
@@ -29,6 +29,10 @@ pub struct Opts {
     #[arg(short = 'H', long)]
     pub hidden: bool,
 
+    /// Silences output
+    #[arg(short = 'Q', long)]
+    pub quiet: bool,
+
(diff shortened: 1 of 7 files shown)


src/opts.rs
```diff
@@ -29,6 +29,10 @@ pub struct Opts {
     #[arg(short = 'H', long)]
     pub hidden: bool,
 
+    /// Silences output
+    #[arg(short = 'Q', long)]
```
marcospb19, line 33: Short flag should be a lowercase 'q'.
```suggestion
    #[arg(short = 'q', long)]
```

marcospb19 (changes requested):

src/opts.rs
a-moreira: you are right! :-)

marcospb19:
quoted from https://github.com/ouch-org/ouch/issues/294#issuecomment-1356976945 
> I wasn't sure if I should send the args.quiet from the main function down the code, like it's done with the question policty and file visibility or not. Let me know what you think pls

I think it's ok to have the argument being passed like this, eventually we'll refactor if it's a problem.

I want to do some local testing here before merging, so I'll be giving another review when possible.

marcospb19 (approved):
I tested it and everything seems to be OK, just a problem with the flag not being global.

```sh
ouch --quiet compress ... # This always works
ouch compress --quiet ... # This only works with 'global = true'
ouch compress ... --quiet # This only works with 'global = true'
```

I'll fix this for --quiet and other flags too, but in another PR.

marcospb19:
Thanks for the PR! :tada: :tada: :tada:

PR #330, opened 2023-01-03
improve extension parsing logic

figsoda:
the two new functions are extracted for the `--format` feature I'm working on

this fixes the issue where `ouch d ..gz` and `ouch d ...gz` tries to delete the current and parent folder respectively, which can be very dangerous with `--yes`, the user will now not be allowed to do that as if the file doesn't have an extension


marcospb19 (approved):
Cool! I'd never noticed that `..gz` and `...gz` could be problematic.

marcospb19:
This fixed an issue a friend had where decompressing `file.cbz` was impossible, because Ouch doesn't detect any known extensions but detects the ZIP format via MIME types, previously it would think the output folder should have the name `file.cbz`, replacing the input by the output.

marcospb19:
I've added this to the CHANGELOG in the bugfixes section: `Fixed overwriting archive file when decompressing archive detected with MIME types.`

PR #332, opened 2023-01-05
refac: use BStr to display possibly non-UTF8 byte sequences

vrmiguel:
Uses `bstr::BStr` to display Paths in the same way that `utils::to_utf` does (without quotes and using the Unicode replacement character) but with less cost: without std's UTF-8 check and without heap allocations when the sequence is non-UTF8

`to_utf` is still used in some places but I aim to remove that in another PR

### DIFF
--- a/src/utils/formatting.rs
+++ b/src/utils/formatting.rs
@@ -1,6 +1,29 @@
-use std::{borrow::Cow, path::Path};
+use std::{borrow::Cow, fmt::Display, path::Path};
+
+use bstr::BStr;
 
 use crate::CURRENT_DIRECTORY;
+
+/// Converts invalid UTF-8 bytes to the Unicode replacement codepoint (�) in its Display implementation.
+pub struct EscapedUtf8Display<'a> {
+    bstr: &'a BStr,
+}
+
+impl<'a> EscapedUtf8Display<'a> {
+    pub fn new(path: &'a Path) -> Self {
+        use std::os::unix::prelude::OsStrExt;
+
+        let bytes = path.as_os_str().as_bytes();
+
+        Self { bstr: BStr::new(bytes) }
+    }
+}
+
+impl Display for EscapedUtf8Display<'_> {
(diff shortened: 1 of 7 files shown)


src/utils/formatting.rs
```diff
@@ -1,7 +1,45 @@
...
+    }
+}
+
+#[cfg(windows)]
+impl Display for EscapedPathDisplay<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        use std::{char, fmt::Write, os::windows::prelude::OsStrExt};
+
+        let utf16 = self.path.as_os_str().encode_wide();
+        let chars = char::decode_utf16(utf16).map(|decoded| decoded.unwrap_or(char::REPLACEMENT_CHARACTER));
```
figsoda, line 33: is this the same as [to_string_lossy](https://doc.rust-lang.org/std/path/struct.Path.html#method.to_string_lossy)?

vrmiguel: it's similar

`to_string_lossy` could be seen as

```rust
fn to_string_lossy(&self) -> Cow<'_, str> {
    if let Some(utf8) = self.to_str() {
         utf8.into()
    } else {
         let utf16 = self.path.as_os_str().encode_wide();
         let chars = char::decode_utf16(utf16)
              .map(|decoded| decoded.unwrap_or(char::REPLACEMENT_CHARACTER));
         chars.collect::<String>().into()
    }
}
```

although since `std` has direct access to `OsStr`'s `Wtf8Buf` they can be smarter at this by using `Wtf8Buf::next_surrogate` as [seen here](https://github.com/rust-lang/rust/blob/master/library/std/src/sys_common/wtf8.rs#L536-L556). that API however is not public so going through UTF-16 decoding seems to be the best option

------

this is preferable to `to_string_lossy` since we manage to avoid the UTF-8 check and the possible String heap allocation when displaying paths

figsoda (approved):
lgtm

PR #334, opened 2023-01-05
Stop keeping track of the names of unpacked files

vrmiguel:
Since we only use the return of `smart_unpack`  to print out the amount of unpacked files, we don't have to build the expensive Vec of unpacked PathBufs, instead only returning the actual number of unpacked files


figsoda (approved):
nice catch

PR #335, opened 2023-01-05
Clean up

figsoda:

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -146,7 +146,7 @@ pub fn run(
                     "You are trying to compress a folder."
                 };
 
-                let error = FinalError::with_title(format!("Cannot compress to '{}'.", output_path))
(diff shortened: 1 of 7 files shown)


src/commands/mod.rs
```diff
@@ -75,10 +75,9 @@ fn check_for_non_archive_formats(files: &[PathBuf], formats: &[Vec<Extension>])
         .iter()
         .zip(formats)
         .filter(|(_, formats)| !formats.first().map(Extension::is_archive).unwrap_or(false))
-        .map(|(path, _)| path)
-        .peekable();
+        .map(|(path, _)| path);
 
-    if not_archives.peek().is_some() {
+    if not_archives.next().is_some() {
```
vrmiguel, line 80: by using `next` we discard the first element of the iterator

i had used `peek` because then the format is still in the iterator to be collected into `not_archives`

figsoda: oops, I overlooked that

PR #337, opened 2023-01-06
Decompress files in parallel

### LINKED ISSUE #244: Decompress files in parallel
We can get big performance improvements by using a threadpool when decompressing files using stuff like Rayon.

[This commit](https://github.com/carmesim/ouch/commit/fe97ff813d2c3a4239f87108ab750b92adb86b59) is an example of how it can be somewhat easily integrable into ouch, but some quirks must be solved:

* error treatment must be adapted 
* threads try to print to stdout simultaneously creating garbled output

vrmiguel:
Uses `rayon` to call `decompress_file` in parallel.

The gain in performance occurs only if more than one file/archive are being decompressed at the same time, e.g. `ouch d videos.tar.gz photos.tar.bz2`. If ouch takes 3 secs to decompress `videos.tar.gz` and 4 secs to decompress `photos.tar.bz2` it's safe to assume it'd take the previous single-threaded code 7 secs to decompress both since the operations take place in sequence. 

Using rayon it's generally safe to assume the runtime is clamped by the time it takes to decompress the biggest file, e.g. it'd multithreaded ouch 4 secs to decompress both `videos.tar.gz` and `photos.tar.bz2`

I guess it's not a very common usecase to decompress many archives at once but I don't think this feature costs much. The binary size, at least, did not seem to increase 

As a next step we can try using [gzp](https://github.com/sstadick/gzp) by @sstadick to use Writers that work in parallel


figsoda (approved):
changes lgtm, but we might want to keep https://github.com/ouch-org/ouch/issues/244 open

marcospb19:
I'm afraid that one of our dependencies can't handle this, idk, previously we had some problems with parallel tests.

Feeling like we should have some tests, @vrmiguel what do you think about re-adding the Python integration tests? We can choose to rewrite it in Rust later, but that's not necessary, I'm just a little annoyed that I don't trust Ouch itself because of the lack of tests.

For example, we don't have any test that does compression of multiple archives at once to make sure this is working, and nobody breaks it in the future.

figsoda:
> we can try using [gzp](https://github.com/sstadick/gzp)

 https://github.com/ouch-org/ouch/pull/348

PR #339, opened 2023-01-06
refactor list command

figsoda:
reduces allocation and locking
`hyperfine` shows about 5% improvement listing the linux git repository with cloned with `--depth 1` as a `.tar` in both the normal and tree mode


marcospb19 (approved):

vrmiguel (approved):
Awesome! 🚀

PR #343, opened 2023-01-10
Add an integration test script

vrmiguel:
Adds a Python integration script (similar to the one we had in 2021)

To-do in this PR:
* [ ] add it to CI
* [ ] add a test for parallel decompression

What other tests would you like to cover? (@marcospb19, @figsoda, other contributors)

### DIFF
--- a/test.py
+++ b/test.py
@@ -1,0 +1,118 @@
+#!/usr/bin/python3
+
+"""
+Integration testing script for Ouch
+"""
+
+import os
+from subprocess import DEVNULL, STDOUT, check_call
+from hashlib import sha3_384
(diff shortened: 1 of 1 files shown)


figsoda:
what does it do that the original tests doesn't?

marcospb19:
I think those were our previous integration tests, I asked @vrmiguel to add the Python script back in this comment https://github.com/ouch-org/ouch/pull/337#issuecomment-1373050212

Maybe this is redundant but we can change it to test the rest of the stuff that we don't have tests yet?

marcospb19:
Closing for now.

PR #345, opened 2023-01-10
Remove incorrect extension Display implementation

marcospb19:
Deleting the extension Display implementation because it assumes
that extensions have one one string equivalent, but some extensions
have multiple aliases.


figsoda:
there is a merge conflict

figsoda (approved):

PR #348, opened 2023-01-31
Multi-threaded compression for gzip and snap using gzp

figsoda:
Some benchmark results (compression):

gzip:
about 10 times the speed for a 12M text file
about 13 times the speed for a lot of regular sized files (tar.gz)
about 10% faster for a lot of small files (tar.gz)

snappy:
about 3 times the speed for a 12M text file
about 5 times the speed for a lot of regular sized files (tar.sz)
**about 20% slower for a lot of small files (tar.sz)**

no difference in size of compressed files, +0.2M in binary size in release mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ Categories Used:
 
 ## [Unreleased](https://github.com/ouch-org/ouch/compare/0.4.1...HEAD)
 
+### Improvements
+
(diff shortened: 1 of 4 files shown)


marcospb19 (changes requested):
Both seem to be good additions, this should probably go to the changelog.

figsoda:
forgot about that

marcospb19 (approved):
Looks good, but I'm a little bit afraid of how much we can trust this library for a new release :sweat_smile: I think I'll politely ask the author.

EDIT: seems like this has been used by sccache.

vrmiguel:
amazing PR 🥺

PR #354, opened 2023-01-31
zip: fix compression of files larger than 4GB

figsoda:
https://github.com/ouch-org/ouch/discussions/353

### DIFF
--- a/src/archive/zip.rs
+++ b/src/archive/zip.rs
@@ -145,7 +145,9 @@ where
     W: Write + Seek,
 {
     let mut writer = zip::ZipWriter::new(writer);
-    let options = zip::write::FileOptions::default();
+    // always use ZIP64 to allow compression of files larger than 4GB
+    // the format is widely supported and the extra 20B is negligible in most cases
+    let options = zip::write::FileOptions::default().large_file(true);
     let output_handle = Handle::from_path(output_path);
 
     #[cfg(not(unix))]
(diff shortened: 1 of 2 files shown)


src/archive/zip.rs
```diff
@@ -219,10 +219,16 @@ where
                     options
                 };
 
-                let mut file = fs::File::open(entry.path())?;
+                let mut file = fs::File::open(path)?;
+                let large = file.metadata().map_or(
+                    true,
+                    |metadata| metadata.len() > 0xffffffff, // 4 GB
```
marcospb19, line 225: Fun fact, `0xffffffff == u32::MAX`.

marcospb19, line 225: Maybe we should just put `true` for every file, `ZIP64` should be very common, what do you think?

From [the documentation](https://docs.rs/zip/latest/zip/write/struct.FileOptions.html#method.large_file) this potentially wastes `20b`, for a thousand files this sums up to `2KB`, which is equivalent to 4 blocks in my filesystem (block size is `512b`), for a thousand-file archive, `2kb` seems irrelevant.

About micro-optimizing this, if we just pick `true`, maybe Rust can do some inlining and remove the branch of dealing with the `false` case.

figsoda: you've convinced me, `File::metadata` seems to require syscalls and the extra 20B might be worth it until we implement multithreaded zip compression

PR #358, opened 2023-02-03
Create `check` module

marcospb19:
This is an attempt of separating the messy error treatment code
from the `run` function, because it was too large.

The new file `check.rs` contains the important usage checks that
are done in runtime.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -27,7 +27,7 @@ use clap::{CommandFactory, ValueEnum};
 use clap_complete::{generate_to, Shell};
 use clap_mangen::Man;
 
-include!("src/opts.rs");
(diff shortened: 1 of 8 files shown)


marcospb19:
5b99f434c3628ad934e43b08bfcb809b626d4a8d and db62f1c53413f60171574f1b610b472d1a5b82d5 probably shouldn't be in this PR :thinking: because it's already big, but I don't think it's a problem.

figsoda (approved):
changes lgtm, though some function names are getting a little long

marcospb19:
Removed vrmiguel's review request because he seems to be busy, I think I can merge this...

PR #361, opened 2023-02-06
make mime sniffing consistent for list and decompress

figsoda:


marcospb19 (approved):
I think this is better, practically a bug fix, I think you don't need to add to the CHANGELOG because `--format` wasn't release, right?

MIME sniffing giving the user questions even after providing `--format` would be insanely annoying.

figsoda:
yeah this shouldn't be in the changelog since `--format` isn't released yet

PR #372, opened 2023-03-14
 add raw level arg

### LINKED ISSUE #52: Add compression level configuration flags
Planned flags:

## `--level LEVEL_NUMBER`
From 1 to 10, `ouch` should "translate" that range for each format, considering that some formats use higher numbers than others.

## `--fast`

Less intensive compressing, we can seek for the best defaults aiming for 50% of time taken (run in half the time).

## `--slow`

More intensive compressing, takes longer, we shall look for defaults that take up to double the time, but also, we'll need to establish a RAM limit.

Some compression methods have a memory usage table that explains how much a certain level can use.

xgdgsc:
partly fix #52 . 

I don' t find the  "translate" necessary as users should know what they are actually using if they are trying to use a compression level.


src/cli/args.rs
```diff
@@ -44,6 +44,10 @@ pub struct CliArgs {
     /// Ouch and claps subcommands
     #[command(subcommand)]
     pub cmd: Subcommand,
+
+    /// Compression raw level as each algo has
+    #[arg(short = 'r', long, global = true, default_value_t = u32::MAX)]
+    pub raw_level: u32,
```
figsoda, line 50: We should use `Option` because `u32::MAX` doesn't mean anything, and we can use things like `map_or_else(Default::default, gzp::par::compress::Compression::new)` and avoid the repetitive `if` statements

`u16` so `as i32` can't panic, the compression levels should be 0-10 anyway

I'm also not sure about the name `--raw-level`, maybe just `--level` or `--compression-level`?

```suggestion
    pub level: Option<u16>,
```

figsoda: as to the clippy error, you can add an `allow`

figsoda: also this is specific to compression and shouldn't be a global flag

marcospb19: Actually, there can be negative compression levels, don't ask me why xD.

https://en.wikipedia.org/wiki/Zstd#:~:text=It%20is%20tunable%20with%20compression,%2C%20but%20best%20compression%20ratio.

figsoda: so i16 would probably make more sense then

marcospb19:
Thanks for your contribution! We really do appreciate it, here's a small change request:

I see you're simulating `Option<u32>` behavior by using `u32::MAX` as a marker for `None`, instead, you can provide `Option<u32>`, `clap` will understand that it is optional and now we'll be able to see in the code that this flag is indeed optional, and maybe wasn't provided.

As Figsoda said, you can also take advantage of this and use `Option`'s methods for default values, instead of doing if-else.

I also think it should be called `--level`.

src/commands/compress.rs
```diff
@@ -44,19 +45,39 @@ pub fn compress_files(
                 // by default, ParCompress uses a default compression level of 3
                 // instead of the regular default that flate2 uses
                 gzp::par::compress::ParCompress::<gzp::deflate::Gzip>::builder()
-                    .compression_level(Default::default())
+                    .compression_level(level.map_or_else(|| Default::default(), |l| gzp::Compression::new(l as u32)))
+                    .from_writer(encoder),
+            ),
+            Bzip => Box::new(bzip2::write::BzEncoder::new(
+                encoder,
+                level.map_or_else(|| Default::default(), |l| bzip2::Compression::new(l as u32)),
```
figsoda, line 53:
```suggestion
                level.map_or_else(Default::default, |l| bzip2::Compression::new(l as u32)),
```

src/commands/compress.rs
```diff
@@ -44,19 +45,39 @@ pub fn compress_files(
...
+            )),
+            Lz4 => Box::new(lzzzz::lz4f::WriteCompressor::new(
+                encoder,
+                lzzzz::lz4f::PreferencesBuilder::new()
+                    .compression_level(level.map_or_else(|| Default::default(), |l| l as i32))
```
figsoda, line 58:
```suggestion
                    .compression_level(level.map_or(0, |l| l as i32))
```

src/commands/compress.rs
```diff
@@ -44,19 +45,39 @@ pub fn compress_files(
...
+                    .build(),
+            )?),
+            Lzma => Box::new(xz2::write::XzEncoder::new(
+                encoder,
+                level.map_or_else(|| Default::default(), |l| l as u32),
```
figsoda, line 63:
```suggestion
                level.map_or(6, |l| l as u32),
```

src/commands/compress.rs
```diff
@@ -44,19 +45,39 @@ pub fn compress_files(
...
-            Lzma => Box::new(xz2::write::XzEncoder::new(encoder, 6)),
-            Snappy => Box::new(gzp::par::compress::ParCompress::<gzp::snap::Snap>::builder().from_writer(encoder)),
             Zstd => {
-                let zstd_encoder = zstd::stream::write::Encoder::new(encoder, Default::default());
+                Box::new(
+                    zstd::stream::write::Encoder::new(encoder, level.map_or_else(|| Default::default(), |l| l as i32))
+                        .unwrap()
+                        .auto_finish(),
+                )
                 // Safety:
```
figsoda, line 78: the comments are no longer valid and also at a wrong position

figsoda (changes requested):

src/commands/compress.rs
```diff
@@ -45,40 +45,41 @@ pub fn compress_files(
...
                 gzp::par::compress::ParCompress::<gzp::deflate::Gzip>::builder()
-                    .compression_level(level.map_or_else(|| Default::default(), |l| gzp::Compression::new(l as u32)))
+                    .compression_level(level.map_or_else(Default::default, |l| gzp::Compression::new(l as u32)))
                     .from_writer(encoder),
             ),
             Bzip => Box::new(bzip2::write::BzEncoder::new(
                 encoder,
-                level.map_or_else(|| Default::default(), |l| bzip2::Compression::new(l as u32)),
+                level.map_or_else(Default::default, |l| {
+                    if l < 1 || l > 9 {
```
figsoda: please use `clamp` instead, the same applies for the other compression methods

tests/integration.rs
```diff
@@ -133,3 +133,23 @@ fn multiple_files(
     ouch!("-A", "d", archive, "-d", after);
     assert_same_directory(before, after, !matches!(ext, DirectoryExtension::Zip));
 }
+
+// compress and decompress a single file with comp levels
+#[proptest(cases = 512)]
+fn single_file_level(
```
figsoda, line 139: this can be part of `single_file` if you make `level` an `Option`

src/commands/compress.rs
figsoda: there is a `clamp` in `std`

src/commands/compress.rs
```diff
@@ -44,19 +45,39 @@ pub fn compress_files(
...
             ),
-            Bzip => Box::new(bzip2::write::BzEncoder::new(encoder, Default::default())),
-            Lz4 => Box::new(lzzzz::lz4f::WriteCompressor::new(encoder, Default::default())?),
-            Lzma => Box::new(xz2::write::XzEncoder::new(encoder, 6)),
-            Snappy => Box::new(gzp::par::compress::ParCompress::<gzp::snap::Snap>::builder().from_writer(encoder)),
             Zstd => {
-                let zstd_encoder = zstd::stream::write::Encoder::new(encoder, Default::default());
+                Box::new(
+                    zstd::stream::write::Encoder::new(encoder, level.map_or_else(|| Default::default(), |l| l as i32))
+                        .unwrap()
```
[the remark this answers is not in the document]
figsoda: please either add the comment back or do error handling

tests/integration.rs
```diff
@@ -100,7 +100,11 @@ fn single_empty_file(ext: Extension, #[any(size_range(0..8).lift())] exts: Vec<F
 
 // compress and decompress a single file
 #[proptest(cases = 512)]
-fn single_file(ext: Extension, #[any(size_range(0..8).lift())] exts: Vec<FileExtension>) {
+fn single_file(
+    ext: Extension,
+    #[any(size_range(0..8).lift())] exts: Vec<FileExtension>,
+    #[strategy(0i16..30)] level: i16,
```
figsoda, line 106:
```suggestion
    #[strategy(0i16..30)] level: Option<i16>,
```
and put the `ouch -A c` line in an `if let` statement

src/commands/compress.rs
```diff
@@ -44,20 +45,39 @@ pub fn compress_files(
...
+            Bzip => Box::new(bzip2::write::BzEncoder::new(
+                encoder,
+                level.map_or_else(Default::default, |l| {
+                    bzip2::Compression::new(num::clamp(l as u32, 1, 9))
+                }),
+            )),
+            Lz4 => Box::new(lzzzz::lz4f::WriteCompressor::new(
+                encoder,
+                lzzzz::lz4f::PreferencesBuilder::new()
+                    .compression_level(level.map_or(0, |l| num::clamp(l as i32, 1, 12)))
```
figsoda, line 62: probably better to use [CLEVEL_MAX](https://docs.rs/lzzzz/latest/lzzzz/lz4f/constant.CLEVEL_MAX.html) here

src/commands/compress.rs
```diff
@@ -44,20 +45,39 @@ pub fn compress_files(
...
+                    .build(),
+            )?),
+            Lzma => Box::new(xz2::write::XzEncoder::new(
+                encoder,
+                level.map_or(6, |l| num::clamp(l as u32, 0, 9)),
+            )),
+            Snappy => Box::new(
+                gzp::par::compress::ParCompress::<gzp::snap::Snap>::builder()
+                    .compression_level(gzp::par::compress::Compression::new(
+                        level.map_or_else(Default::default, |l| num::clamp(l as u32, 0, 9)),
```
figsoda, line 72: this should be consistent with the other gzp compression, either use `0..=9` or `1..=9`

tests/integration.rs
xgdgsc: Didn' t find how to the strategy with Option.

src/commands/compress.rs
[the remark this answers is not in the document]
xgdgsc: Don' t know how to handle error here. Didn' t the comment say the only possible error is level range?

tests/integration.rs
figsoda: this will work
```suggestion
    #[strategy(proptest::option::of(0i16..30))] level: Option<i16>,
```

and you would need to change the below to this
```rust
    if let Some(level) = level {
        ouch!("-A", "c", "-l", level.to_string(), before_file, archive);
    } else {
        ouch!("-A", "c", before_file, archive);
    }
```

src/commands/compress.rs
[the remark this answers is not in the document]
figsoda: then the comment should be brought back

src/commands/compress.rs
```diff
@@ -44,20 +45,42 @@ pub fn compress_files(
...
+                gzp::par::compress::ParCompress::<gzp::snap::Snap>::builder()
+                    .compression_level(gzp::par::compress::Compression::new(
+                        level.map_or_else(Default::default, |l| (l as u32).clamp(0, 9)),
+                    ))
+                    .from_writer(encoder),
+            ),
+            Zstd => Box::new(
+                zstd::stream::write::Encoder::new(
+                    encoder,
+                    level.map_or(0, |l| {
```
marcospb19, line 77: here you can use `zstd::DEFAULT_COMPRESSION_LEVEL` directly, it's 3 instead of 0.
```suggestion
                    level.map_or(zstd::DEFAULT_COMPRESSION_LEVEL, |l| {
```

marcospb19 (changes requested):
I reviewed the compression levels for each format and three of them need adjustment:

src/commands/compress.rs
```diff
@@ -44,20 +45,42 @@ pub fn compress_files(
...
-                Box::new(zstd_encoder.unwrap().auto_finish())
-            }
+            Bzip => Box::new(bzip2::write::BzEncoder::new(
+                encoder,
+                level.map_or_else(Default::default, |l| bzip2::Compression::new((l as u32).clamp(1, 9))),
+            )),
+            Lz4 => Box::new(lzzzz::lz4f::WriteCompressor::new(
+                encoder,
+                lzzzz::lz4f::PreferencesBuilder::new()
+                    .compression_level(level.map_or(0, |l| (l as i32).clamp(1, lzzzz::lz4f::CLEVEL_MAX)))
```
marcospb19, line 60: This one is weird, [lzzz](https://docs.rs/lzzzz/latest/src/lzzzz/lz4f/preferences.rs.html#4-5) shows a default of `0`, but in `man lz4` the default is `1`, also, the range starts at `1`, so `0` doesn't make much sense.

Maybe it's a difference between lz4f and lz4? Idk

You're clamping with 1 so I guess this does make sense:

```suggestion
                    .compression_level(level.map_or(1, |l| (l as i32).clamp(1, lzzzz::lz4f::CLEVEL_MAX)))
```

src/commands/compress.rs
```diff
@@ -44,20 +45,42 @@ pub fn compress_files(
...
-            Zstd => {
-                let zstd_encoder = zstd::stream::write::Encoder::new(encoder, Default::default());
-                // Safety:
-                //     Encoder::new() can only fail if `level` is invalid, but Default::default()
-                //     is guaranteed to be valid
-                Box::new(zstd_encoder.unwrap().auto_finish())
-            }
+            Bzip => Box::new(bzip2::write::BzEncoder::new(
+                encoder,
+                level.map_or_else(Default::default, |l| bzip2::Compression::new((l as u32).clamp(1, 9))),
```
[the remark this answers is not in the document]
xgdgsc: changing to 0,9 cause test failure

src/commands/compress.rs
```diff
@@ -44,15 +45,39 @@ pub fn compress_files(
...
-                let zstd_encoder = zstd::stream::write::Encoder::new(encoder, Default::default());
+                let zstd_encoder = zstd::stream::write::Encoder::new(
+                    encoder,
+                    level.map_or(zstd::DEFAULT_COMPRESSION_LEVEL, |l| {
+                        (l as i32).clamp(zstd::zstd_safe::min_c_level(), zstd::zstd_safe::max_c_level())
+                    }),
+                );
                 // Safety:
                 //     Encoder::new() can only fail if `level` is invalid, but Default::default()
                 //     is guaranteed to be valid
```
figsoda, line 83:
```suggestion
                //     Encoder::new() can only fail if `level` is invalid, but the level
                //     is `clamp`ed and therefore guaranteed to be valid
```

figsoda (approved):

figsoda:
you might also want to allow `clippy::too-many-arguments` just to make clippy happy

src/commands/compress.rs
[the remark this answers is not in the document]
marcospb19: ok, resolving it for now.

marcospb19 (approved):
Looks great, thanks!

PR #374, opened 2023-03-19
implement `--fast` and `--slow`

### LINKED ISSUE #373: Add flags for slowest and fastest compression levels
(follow up to #52)

For convenience, we could have a flag for the fastest and slowest compression levels.

They must conflict with each other and the newly added `--level`.

Implementation details: you can use `i16::MAX` and `i16::MIN` because levels are always clamped to the min-max range.

figsoda:
closes #373

should `--fast` be 1 instead of 0?

### DIFF
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -61,8 +61,18 @@ pub enum Subcommand {
         output: PathBuf,
 
         /// Compression level, applied to all formats
-        #[arg(short, long)]
+        #[arg(short, long, group = "compression-level")]
         level: Option<i16>,
+
+        /// Fastest compression level possible,
+        /// conflicts with --level and --slow
+        #[arg(long, group = "compression-level")]
+        fast: bool,
+
+        /// Slowest (and best) compression level possible,
+        /// conflicts with --level and --fast
+        #[arg(long, group = "compression-level")]
+        slow: bool,
     },
     /// Decompresses one or more files, optionally into another folder
     #[command(visible_alias = "d")]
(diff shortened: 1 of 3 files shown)


src/cli/args.rs
```diff
@@ -61,8 +61,18 @@ pub enum Subcommand {
         output: PathBuf,
 
         /// Compression level, applied to all formats
-        #[arg(short, long)]
+        #[arg(short, long, group = "compression-level")]
         level: Option<i16>,
+
+        /// Fastest compression level possible,
+        /// conflicts with --level and --slow
```
marcospb19, line 68: I think it's not necessary to add the conflict message to --help, what do you think?

src/commands/mod.rs
```diff
@@ -81,7 +83,13 @@ pub fn run(
...
                 question_policy,
                 file_visibility_policy,
-                level,
+                if fast {
+                    Some(0)
+                } else if slow {
+                    Some(i16::MAX)
+                } else {
+                    level
+                },
```
marcospb19, line 92: Can you bind this expression to a variable? Nested expressions are one of my strong pet peeves xD, I like giving names because `Some(0)` doesn't have a clear meaning.

marcospb19 (changes requested):

marcospb19:
> should --fast be 1 instead of 0

oh yeah, I belive so, people will expect `--fast` to still do some compression!

otherwise I think `--fast` would be kinda useless

src/cli/args.rs
figsoda: it is in `--help`

```
      --fast             Fastest compression level possible, conflicts with --level and --slow
      --slow             Slowest (and best) compression level possible, conflicts with --level and --fast
```

marcospb19 (approved):

PR #376, opened 2023-03-19
set up dependabot

figsoda:
or maybe we should use renovate instead?


marcospb19 (approved):
I don't have a strong opinion about renovate vs dependabot. :+1:

PR #379, opened 2023-03-20
tests: reduce maximum compression level

figsoda:
CI sometimes runs out of memory on cross, this should ease the issue a bit


marcospb19 (approved):
To be honest I didn't spotted the CI runs where this issue happened, I just cancelled the last run because it was a README change.

I'd say we should probably use a lower value to affect the formats that take the most time, like .xz and .xz.

however, I tried searching for how much zstd is actually using and didn't find a table of values, so idk.

figsoda:
lowering the maximum lowers the memory consumption overall since we will now have a lower average and such

Issue #293, opened 2022-10-15
Tar archives are bigger than expected (when containing symlinks)


marcospb19:
### Version

bda6c270629711cd9072826471553c58766daf56 (main branch)

### Description

Created `.tar` and `.tar.gz` archives are bigger than normal.

### Current Behavior

Compressing the following folder with `447 MB`:

```
$ du nvidia-utils -hs
447M    nvidia-utils
```

With `Ouch`:
```
$ ouch compress nvidia-utils ouch.tar
```

And `Tar`:
```
$ tar -cvf tar.tar nvidia-utils
```

Produces two archives with different sizes:
```
$ du *.tar -hs --apparent-size
1.1G    ouch.tar
447M    tar.tar
```

### Expected Behavior

The size of the archive produced by Ouch should be similar to the size of the folder used as input.

marcospb19:
@figsoda guessed the reason, so `Ouch` is following symlinks and copying the pointed files, instead of copying the symlink itself to the archive.

We should probably copy symlinks too by default, it seems more predictable.

This is related to #232.

marcospb19:
This is blocked because I could not figure out why the Tar library is panicking when I try to add symlinks to it.

PR #392, opened 2023-04-01
`tar`: store symlinks in archive

marcospb19:
TODO


marcospb19:
This was an attempt at "fixing" https://github.com/ouch-org/ouch/issues/293 so that archives can't become unnecessarily huge, but I couldn't make this work, closing for now.

PR #432, opened 2023-06-03
decompress: don't panic on `.tar.zip`, `.zip.zip`, etc

### LINKED ISSUE #431: Duplicating file suffix causes panic: "entered unreachable code"
### Version

0.4.1

### Description

I came a cross this panic when I was trying to decompress a ZIP file with two `.zip` suffixes. It's easy to reproduce:

image

### Current Behavior

Decompressing a file like `test.zip.zip` with two suffixes causes the program to panic.

### Expected Behavior

The program should decompress the file normally and guess the file's type using the last suffix.

### Additional Information

_No response_

vrmiguel:
Hopefully closes #431 

See https://github.com/ouch-org/ouch/issues/431#issuecomment-1575100804 for context

### DIFF
--- a/src/commands/list.rs
+++ b/src/commands/list.rs
@@ -52,7 +52,7 @@ pub fn list_archive_contents(
                 Lzma => Box::new(xz2::read::XzDecoder::new(decoder)),
                 Snappy => Box::new(snap::read::FrameDecoder::new(decoder)),
                 Zstd => Box::new(zstd::stream::Decoder::new(decoder)?),
-                Tar | Zip => unreachable!(),
+                Tar | Zip => decoder,
             };
             Ok(decoder)
         };
--- a/src/commands/decompress.rs
+++ b/src/commands/decompress.rs
@@ -86,7 +86,7 @@ pub fn decompress_file(
             Lzma => Box::new(xz2::read::XzDecoder::new(decoder)),
             Snappy => Box::new(snap::read::FrameDecoder::new(decoder)),
             Zstd => Box::new(zstd::stream::Decoder::new(decoder)?),
-            Tar | Zip => unreachable!(),
+            Tar | Zip => decoder,
         };
         Ok(decoder)
     };


figsoda (approved):

marcospb19:
Previous behavior: we always decompress all formats in the chain for all supported files.

This PR breaks this, I'm clueless what are the implications of this.

We could suggest `ouch decompress archive.zip.zip --format zip` when we see more than one archive format in the format chain?

Would this be better or worse than the approach here? What do you guys think?

src/commands/list.rs
```diff
@@ -52,7 +52,7 @@ pub fn list_archive_contents(
                 Lzma => Box::new(xz2::read::XzDecoder::new(decoder)),
                 Snappy => Box::new(snap::read::FrameDecoder::new(decoder)),
                 Zstd => Box::new(zstd::stream::Decoder::new(decoder)?),
-                Tar | Zip => unreachable!(),
+                Tar | Zip => decoder,
```
marcospb19, line 55: I'd like a test to make sure we don't break this in the future.

Where it asserts that:

```
.zip.zip.zip
.zip.zip
.zip.tar
.tar.tar.tar
.tar.tar
.tar.zip
```

All generate a file with the correct path name.

marcospb19 (changes requested):

src/commands/list.rs
vrmiguel: I tried but I'm blocked since `ouch` does not yet allow compressing into something like `.zip.zip` :cry: 

We'd have to do that first before making this test

marcospb19: @vrmiguel it's possible
```sh
$ ouch compress README.md archive.zip
$ ouch compress archive.zip archive.zip.zip --format zip
```

marcospb19:
(I deleted an outdated comment of mine and it had a thread so it shows like I deleted vrmiguel's. Should've just collapsed it instead.)

Closing in favor of #817.

PR #433, opened 2023-06-04
fix: handle Zip when modified times are missing

### LINKED ISSUE #411: Crash when files in ZIP have no date
### Version

0.4.1

### Description

When trying to decompress a ZIP that contains files with no modifying and/or creation date, `ouch` crashes.

### Current Behavior

`ouch` crashes with the following message

> [INFO] Created temporary directory /mnt/c/Users/jonas.frei/Documents/tmp/dl/./.tmpPENU32 to hold decompressed elements.
> [INFO] ".tmpPENU32/ch6f_34080_091a_1.dxf" extracted. (313.98KiB)
> thread 'main' panicked at 'Zip archive contains a file with broken 'last modified time': ComponentRange { name: "month", minimum: 1, maximum: 12, value: 0, conditional_range: false }', src/archive/zip.rs:267:10
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

### Expected Behavior

Shouldn't crash, use current datetime

### Additional Information

I think usually the files in a ZIP do have this metadata. Those ZIP files were generated by a tool that apparently doesn't include this information.

marcospb19:
Fixes #411.

### DIFF
--- a/src/archive/zip.rs
+++ b/src/archive/zip.rs
@@ -261,14 +261,14 @@ fn get_last_modified_time(file: &fs::File) -> DateTime {
 }
 
 fn set_last_modified_time(zip_file: &ZipFile, path: &Path) -> crate::Result<()> {
-    let modification_time_in_seconds = zip_file
-        .last_modified()
-        .to_time()
-        .expect("Zip archive contains a file with broken 'last modified time'")
-        .unix_timestamp();
+    let modification_time = zip_file.last_modified().to_time();
+
+    let Ok(time_in_seconds) = modification_time else {
+        return Ok(());
+    };
 
     // Zip does not support nanoseconds, so we can assume zero here
-    let modification_time = FileTime::from_unix_time(modification_time_in_seconds, 0);
+    let modification_time = FileTime::from_unix_time(time_in_seconds.unix_timestamp(), 0);
 
     set_file_mtime(path, modification_time)?;
 


src/archive/zip.rs
```diff
@@ -261,14 +261,15 @@ fn get_last_modified_time(file: &fs::File) -> DateTime {
...
 
 fn set_last_modified_time(zip_file: &ZipFile, path: &Path) -> crate::Result<()> {
-    let modification_time_in_seconds = zip_file
-        .last_modified()
-        .to_time()
-        .expect("Zip archive contains a file with broken 'last modified time'")
-        .unix_timestamp();
+    let modification_time = zip_file.last_modified().to_time();
+
+    let time_in_seconds = match modification_time {
```
figsoda, line 266: btw you can use `let else` here

figsoda (approved):
lgtm

src/archive/zip.rs
marcospb19: First I did it with `let-else`, but it's that use-case where I start questioning my rustfmt gut. 

```rust
let modification_time = zip_file
    .last_modified()
    .to_time()
else {
    return Ok(());
};
```

:eyes: do you like this?

figsoda: I think rustfmt just gave up since it is not implemented, and you can manually format it however you want

You can separate it into 2 statements like what you have right now if that makes it easier to format manually

marcospb19: Oh, I see what you mean, will do and merge.

marcospb19:
oh, CI is failing

figsoda:
the failure seems unrelated, I think we are just running into edge case upstream bugs

PR #445, opened 2023-06-20
switch to lz4_flex

PSeitz:
The release of lz4_flex 0.11 fixes two issue that caused problems
before:
- The new `auto-finish` flushes on drop
- FrameEncoder writes now empty frames. Previously nothing was written,
  which caused issues with double compressed empty data.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -686,6 +686,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "lz4_flex"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ad6f50e0fd4351da4b429de1234c28feda97aafa5748fbd4a74634f5cc5d82dd"
+dependencies = [
+ "twox-hash",
+]
+
+[[package]]
 name = "lzma-sys"
 version = "0.1.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


marcospb19 (approved):
Sorry for the delay.

I did some benchmarks before merging.

The more I chain decoders, the better results I get:

1. `.lz4` -> 1.3x
2. `.lz4.lz4` -> 1.6x
4. `.lz4.lz4.lz4.lz4` -> 2.0x

```
Summary
  ./ouch-new compress input.txt output.lz4.lz4 ran
    1.63 ± 0.02 times faster than ./ouch-previous compress input.txt output.lz4.lz4
```

I also saw improvements when decompressing.

Input file was generated with the following command:
```
dd if=/dev/random of=input.txt bs=1M count=500
```

All tests done in `/tmp`, because it is a filesystem that lives in-memory (doesn't add disk IO to the mix).

Benchmark command used:
```
hyperfine -w 3 -p 'rm output.lz4.lz4' './ouch-previous compress input.txt output.lz4.lz4' './ouch-new compress input.txt output.lz4.lz4'
```

In my machine (using `nightly-x86_64-unknown-linux-gnu`), I also saw a reduction in binary size of 123kb (from 3589 to 3466).

Thanks for your contribution!

figsoda:
Thank you for benchmarking this @marcospb19, and thank you for implementing this and fixing the bugs @PSeitz! I didn't have to run the benchmarks initially and kinda forgot about it (sorry), this looks great!

PR #508, opened 2023-09-12
completions: hint generator to expand to file paths

marcospb19:


src/cli/args.rs
```diff
@@ -78,18 +78,18 @@ pub enum Subcommand {
...
     Decompress {
         /// Files to be decompressed
-        #[arg(required = true, num_args = 1..)]
+        #[arg(required = true, num_args = 1.., value_hint = ValueHint::FilePath)]
         files: Vec<PathBuf>,
 
         /// Place results in a directory other than the current one
-        #[arg(short = 'd', long = "dir", value_hint = ValueHint::DirPath)]
+        #[arg(short = 'd', long = "dir", value_hint = ValueHint::FilePath)]
         output_dir: Option<PathBuf>,
```
marcospb19, line 86: There are some times I want the `-d` flag to expand to a file name, like

```sh
ouch decompress my-pictures-with-family-193-123-12-581.tar.gz -d <TAB-HERE>
```

I'd like it to match `my-pictures-with-family-193-123-12-581.tar.gz`, so I can erase some characters and end up with `my-pictures-with-family`.

figsoda (approved):
Didn't know this would work with completions!

Issue #485, opened 2023-08-16
`--format` flag order matters


marcospb19:
### Version

0.4.2

### Description

When compressing an archive with `--format`, not all positions are supported and the error message isn't helpful.

### Current Behavior

```sh
ouch compress a b c output --format tar.gz # 1
ouch compress a b c --format tar.gz output # 2
ouch compress a b --format tar.gz c output # 3
ouch compress a --format tar.gz b c output # 4
ouch compress --format tar.gz a b c output # 5
```

1 and 5 works, 2-4 doesn't.

Error message for `3`:

```
$ ouch compress a b --format tar.gz c output
error: unexpected argument 'c' found

Usage: ouch compress [OPTIONS] <FILES>... <OUTPUT>

For more information, try '--help'.
```

---

To show that it also happens with a single file as input:

```sh
ouch compress input output --format gz # 6
ouch compress input --format gz output # 7
ouch compress --format gz input output # 8
```

6 and 8 works, 7 doesn't.

Error message for `7`:

```
$ ouch compress input --format gz output
error: unexpected argument 'output' found

Usage: ouch compress [OPTIONS] <FILES>... <OUTPUT>

For more information, try '--help'.
```

### Expected Behavior

1-8 should all work.

### Additional Information

We should add more tests for CLI usage that cover these.

marcospb19:
That's actually a current limitation of clap, issue at https://github.com/clap-rs/clap/issues/1125.

PR #511, opened 2023-09-12
add tests for CLI usage

marcospb19:

### DIFF
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -8,7 +8,7 @@ use clap::{Parser, ValueHint};
 /// Supported formats: tar, zip, gz, xz/lzma, bz/bz2, lz4, sz, zst.
 ///
 /// Repository: https://github.com/ouch-org/ouch
-#[derive(Parser, Debug)]
+#[derive(Parser, Debug, PartialEq)]
 #[command(about, version)]
 // Disable rustdoc::bare_urls because rustdoc parses URLs differently than Clap
 #[allow(rustdoc::bare_urls)]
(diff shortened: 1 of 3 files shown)


figsoda:
Aren't we already testing most of them in the integration tests?

marcospb19:
Probably, but it didn't cover #485 case, I wrote these tests while trying to solve that issue.

figsoda:
yeah sounds good to me then

src/cli/args.rs
```diff
@@ -97,3 +97,158 @@ pub enum Subcommand {
         tree: bool,
     },
 }
+
+#[cfg(test)]
+impl Default for CliArgs {
```
figsoda, line 102: I think we can just put this in a function inside `mod tests`

marcospb19:
I don't plan on removing your property tests, they're useful and also provide us some fuzzing, but I might add some specific tests in the future (the objective is not to be redundant, but I might accidentally get redundant sometimes).

Talking about our proptests, can you tell why this run failed?

https://github.com/ouch-org/ouch/actions/runs/6113846588/job/16594167075

Mmmmm that's the bad thing about only having tests with a scope so broad, investigating can be very confusing.

figsoda:
From the error message, it looks like the runner might have run out of memory possibly due to the fact that we are using cross and the combination of compression methods happened to take more memory than usual

marcospb19:
> From the error message, it looks like the runner might have run out of memory possibly due to the fact that we are using cross and the combination of compression methods happened to take more memory than usual

That's great to hear :D , do you suggest we do something in order to reduce the chance of getting another false-positive?

figsoda:
Maybe trying to decrease the number of extensions, or decrease the maximum size of the file

src/cli/args.rs
marcospb19: :+1: changed and force-pushed.

figsoda (approved):

PR #514, opened 2023-09-15
make CI faster by decreasing size of tests

marcospb19:
make fuzzing tests create less files, with half the size and run less test cases

should speed up by a magnitude of ~6.0x


marcospb19:
this should probably solve out-of-ordinary memory consumption CI failures like the one mentioned at https://github.com/ouch-org/ouch/pull/511#issuecomment-1716523707.

marcospb19:
I'll merge for now.

PR #519, opened 2023-09-18
Refac + Improve extension parsing

### LINKED ISSUE #517: Error message for `--format .tar.gz` outputs empty extension
### Version

0.4.2

### Description

When running
```sh
ouch compress LICENSE output --format .tar.gz
```

### Current Behavior

ouch outputs

```
[ERROR] Invalid archive format
 - Unsupported extension:
```

### Expected Behavior

Properly handle the `--format` flag with the leading dot.

OR

Output this instead: 

```
[ERROR] Invalid archive format
 - Unsupported extension: .tar.gz
```

marcospb19:
Fixes #517.


src/extension.rs
```diff
@@ -123,17 +120,30 @@ fn split_extension(name: &mut &[u8]) -> Option<Extension> {
...
+                reason: format!("Unsupported extension '{}'", extension),
+            })
+        })
+        .collect::<crate::Result<_>>()?;
+
+    if extensions.is_empty() {
+        return Err(Error::InvalidFormatFlag {
+            text: input.to_owned(),
+            reason: "Parsing got an empty list of extensions.".to_string(),
+        });
```
marcospb19, line 146: This is the main change of this PR.

src/extension.rs
```diff
@@ -235,6 +245,69 @@ mod tests {
...
+        );
+        assert_eq!(
+            separate_known_extensions_from_name(".tar.gz".as_ref()),
+            (".tar".as_ref(), vec![Extension::new(&[Gzip], "gz")])
+        );
+    }
+
+    #[test]
+    /// Test extension parsing of `--format FORMAT`
+    fn test_parse_of_format_flag() {
```
marcospb19, line 282: Added tests to cover `--format`.

src/extension.rs
```diff
@@ -235,6 +245,69 @@ mod tests {
         assert_eq!(formats, vec![Tar, Gzip]);
     }
 
+    #[test]
+    /// Test extension parsing for input/output files
+    fn test_separate_known_extensions_from_name() {
```
marcospb19, line 250: As an extra, added tests for extension parsing for input and output files.

tests/snapshots/ui__ui_test_err_format_flag-3.snap
```diff
@@ -0,0 +1,15 @@
...
+ - Unsupported extension '$#!@'
+
+hint: Supported extensions are: tar, zip, bz, bz2, gz, lz4, xz, lzma, sz, zst
+hint: Supported aliases are: tgz, tbz, tlz4, txz, tzlma, tsz, tzst
+hint: 
+hint: Examples:
+hint:   --format tar
+hint:   --format gz
+hint:   --format tar.gz
+
```
marcospb19, line 15: This is the new error message

tests/ui.rs
```diff
@@ -87,6 +87,29 @@ fn ui_test_err_missing_files() {
...
+fn ui_test_ok_format_flag() {
+    let (_dropper, dir) = testdir().unwrap();
+
+    // prepare
+    run_in(dir, "touch", "input").unwrap();
+
+    ui!(run_ouch("ouch compress input output1 --format tar.gz", dir));
+    ui!(run_ouch("ouch compress input output2 --format .tar.gz", dir));
+}
+
```
marcospb19, line 112: The new UI tests to cover `--format`.

marcospb19 (commented):
One thing led to another and I ended up making lots of changes unrelated to the fix.

So are the main changes of this PR:

figsoda:
this seems like a lot, could this be separated into multiple PRs? (or maybe reword the PR to reflect all the changes?)

src/extension.rs
```diff
@@ -123,17 +120,28 @@ fn split_extension(name: &mut &[u8]) -> Option<Extension> {
     Some(ext)
 }
 
-pub fn parse_format(fmt: &OsStr) -> crate::Result<Vec<Extension>> {
-    let fmt = <[u8] as ByteSlice>::from_os_str(fmt).ok_or_else(|| Error::InvalidFormat {
-        reason: "Invalid UTF-8".into(),
+pub fn parse_format_flag(input: &OsStr) -> crate::Result<Vec<Extension>> {
```
figsoda, line 123: I think we can just strip the leading `.` here?

src/extension.rs
marcospb19: @figsoda I basically added this line:
```
.filter(|extension| !extension.is_empty())
```

Which covers the case of stripping the leading `.`.

marcospb19:
Yeah sorry for that, I'll split the PR into smaller ones.

marcospb19:
I gave up on separating this into smaller PRs, I'll avoid doing this again, but I did a mess and it would take effort to separate.

So... merging :v (when it passes).

src/extension.rs
marcospb19: We could, but I already did this code that makes it even "more forgiving", while unrealistic that someone will type ".tar..gz" I also don't see a problem with it so I'll keep it.

PR #522, opened 2023-09-20
Add support for bzip3

### LINKED ISSUE #398: Add support for bzip3
See : https://github.com/kspalaiologos/bzip3
Also see: https://crates.io/crates/bzip3

freijon:
Closes #398
Signed-off-by: Jonas Frei <freijon@pm.me>

This PR adds support for the [bzip3](https://github.com/kspalaiologos/bzip3) format. I tested it to the best of my knowledge and it seemed to work fine. Unlike the other formats, one can not modify the compression level directly. Bzip3 works with block sizes which can be varied. I chose a block size of 5MB which yielded good results with most examples I tested.

### ⚠️ Things to consider before merging
- The project is quite young (May 2022)
- While their name is bzip3, they have no affiliation with the original bz/bz2 and I don't know if they have the permission of the original author of bz/bz2 to use this name. Nevertheless, a lot of popular Linux distros added the library to [their repository](https://repology.org/project/bzip3/versions). 
- The Rust crate [bzip3-rs](https://github.com/bczhc/bzip3-rs) uses the unstable `try` block and may cause complications at build time

P.S. I'm a Rust noob and did this primarily as a proof of concept. Let me know what you think about this.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -122,7 +122,6 @@ jobs:
             target/${{ matrix.target }}/release/ouch.exe
             artifacts/
 
-
   clippy-rustfmt:
     name: clippy-rustfmt
     runs-on: ubuntu-latest
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ atty = "0.2.14"
 bstr = { version = "1.10.0", default-features = false, features = ["std"] }
 bytesize = "1.3.0"
 bzip2 = "0.4.4"
-bzip3 = { version = "0.8.3", features = ["bundled"] }
+bzip3 = { version = "0.9.0", features = ["bundled"] }
 clap = { version = "4.5.20", features = ["derive", "env"] }
 filetime_creation = "0.2"
 flate2 = { version = "1.0.30", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ atty = "0.2.14"
 bstr = { version = "1.10.0", default-features = false, features = ["std"] }
 bytesize = "1.3.0"
 bzip2 = "0.4.4"
-bzip3 = "0.8.1"
+bzip3 = { git = "https://github.com/bczhc/bzip3-rs", rev = "094b80c321732636f32733cc72f05ac3deffbe88", features = ["bundled"] }
 clap = { version = "4.5.20", features = ["derive", "env"] }
 filetime_creation = "0.2"
 flate2 = { version = "1.0.30", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ atty = "0.2.14"
 bstr = { version = "1.10.0", default-features = false, features = ["std"] }
 bytesize = "1.3.0"
 bzip2 = "0.4.4"
-bzip3 = { git = "https://github.com/bczhc/bzip3-rs", rev = "094b80c321732636f32733cc72f05ac3deffbe88", features = ["bundled"] }
+bzip3 = { version = "0.8.3", features = ["bundled"] }
 clap = { version = "4.5.20", features = ["derive", "env"] }
 filetime_creation = "0.2"
 flate2 = { version = "1.0.30", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,9 @@ description = "A command-line utility for easily compressing and decompressing f
 [dependencies]
 atty = "0.2.14"
 bstr = { version = "1.10.0", default-features = false, features = ["std"] }
+bytesize = "1.3.0"
 bzip2 = "0.4.4"
+bzip3 = "0.8.1"
 clap = { version = "4.5.20", features = ["derive", "env"] }
 filetime_creation = "0.2"
 flate2 = { version = "1.0.30", default-features = false }
--- a/src/commands/compress.rs
+++ b/src/commands/compress.rs
@@ -97,7 +97,7 @@ pub fn compress_files(
     match first_format {
         Gzip | Bzip | Bzip3 | Lz4 | Lzma | Snappy | Zstd => {
             writer = chain_writer_encoder(&first_format, writer)?;
-            let mut reader = fs::File::open(&files[0]).unwrap();
+            let mut reader = fs::File::open(&files[0])?;
 
             io::copy(&mut reader, &mut writer)?;
         }
--- a/src/commands/compress.rs
+++ b/src/commands/compress.rs
@@ -57,8 +57,8 @@ pub fn compress_files(
                 level.map_or_else(Default::default, |l| bzip2::Compression::new((l as u32).clamp(1, 9))),
             )),
             Bzip3 => Box::new(
-                // Unwrap is safe when using a valid block size
-                bzip3::write::Bz3Encoder::new(encoder, bytesize::ByteSize::mib(5).0 as usize).unwrap(),
+                // Use block size of 16 MiB
+                bzip3::write::Bz3Encoder::new(encoder, 16 * 2_usize.pow(20))?,
             ),
             Lz4 => Box::new(lz4_flex::frame::FrameEncoder::new(encoder).auto_finish()),
             Lzma => Box::new(xz2::write::XzEncoder::new(
--- a/src/commands/compress.rs
+++ b/src/commands/compress.rs
@@ -56,6 +56,10 @@ pub fn compress_files(
                 encoder,
                 level.map_or_else(Default::default, |l| bzip2::Compression::new((l as u32).clamp(1, 9))),
             )),
+            Bzip3 => Box::new(
+                // Unwrap is safe when using a valid block size
+                bzip3::write::Bz3Encoder::new(encoder, bytesize::ByteSize::mib(5).0 as usize).unwrap(),
+            ),
             Lz4 => Box::new(lz4_flex::frame::FrameEncoder::new(encoder).auto_finish()),
             Lzma => Box::new(xz2::write::XzEncoder::new(
                 encoder,
@@ -91,7 +95,7 @@ pub fn compress_files(
     }
 
     match first_format {
-        Gzip | Bzip | Lz4 | Lzma | Snappy | Zstd => {
+        Gzip | Bzip | Bzip3 | Lz4 | Lzma | Snappy | Zstd => {
             writer = chain_writer_encoder(&first_format, writer)?;
             let mut reader = fs::File::open(&files[0]).unwrap();
 
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -21,6 +21,7 @@ enum DirectoryExtension {
     Tar,
     Tbz,
     Tbz2,
+    Tbz3,
     Tgz,
     Tlz4,
     Tlzma,
@@ -36,6 +37,7 @@ enum DirectoryExtension {
 enum FileExtension {
     Bz,
     Bz2,
+    Bz3,
     Gz,
     Lz4,
     Lzma,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Categories Used:
 ### New Features
 
 - Add multithreading support for `zstd` compression [\#689](https://github.com/ouch-org/ouch/pull/689) ([nalabrie](https://github.com/nalabrie))
+- Add `bzip3` support [\#522](https://github.com/ouch-org/ouch/pull/522) ([freijon](https://github.com/freijon))
 
 ### Bug Fixes
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -256,9 +256,9 @@ dependencies = [
 
 [[package]]
 name = "bzip3"
-version = "0.8.3"
+version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a396e70a91080ca308fe4d37ebf2d15b444beef3b53853d9e36110b3a331e82e"
+checksum = "e908c8163856f1138894e7bce70d2ae79c2d9a19a515dfe96e10888a839aa10c"
 dependencies = [
  "byteorder",
  "bytesize",
@@ -831,9 +831,9 @@ checksum = "830d08ce1d1d941e6b30645f1a0eb5643013d835ce3779a5fc208261dbe10f55"
 
 [[package]]
 name = "libbzip3-sys"
-version = "0.3.3+1.3.2"
+version = "0.4.0+1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1fb2f193373a540601f16a3375d351d685e4ca288c766fe35ce035874c2711f1"
+checksum = "001a0e4d146f247a741755ebc538d592b3603923df85d6e4472e4268822b5d6b"
 dependencies = [
  "bindgen",
  "cc",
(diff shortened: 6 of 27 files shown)


marcospb19:
Good job!

I have a branch with this implemented as an experiment, sort of, https://github.com/ouch-org/ouch/compare/main...bzip3 our code looks very similar.

I've been busy so I'm really glad to have your help.

## Previous Context

I encountered two blockers:

1. Licensing doubts solved in https://github.com/bczhc/bzip3-rs/issues/1.
2. Ouch tests failing, hopefully :crossed_fingers: solved addressed by this PR I made https://github.com/bczhc/bzip3-rs/pull/4, as the maintainer quickly followed up with a fix.

## New Context

Seems like you just found a new blocker:

3. Our releases are built in `stable Rust`, I might have to kindly ask `bzip3-rs`'s maintainer to remove the `try_blocks` used in the build script. (PRs open: https://github.com/bczhc/bzip3-rs/pull/6 or https://github.com/bczhc/bzip3-rs/pull/7).

If you want to attempt making CI pass while there's no release of `bzip3-rs`, make the following change in `Cargo.toml`:
```diff
- bzip3 = "0.8.1"
+ bzip3 = { git = "https://github.com/marcospb19/bzip3-rs", branch = "make-build-script-run-in-stable-1" }
```

src/commands/compress.rs
```diff
@@ -55,6 +55,10 @@ pub fn compress_files(
                 encoder,
                 level.map_or_else(Default::default, |l| bzip2::Compression::new((l as u32).clamp(1, 9))),
             )),
+            Bzip3 => Box::new(
+                // Unwrap is safe when using a valid block size
+                bzip3::write::Bz3Encoder::new(encoder, bytesize::ByteSize::mib(5).0 as usize).unwrap(),
+            ),
```
marcospb19, line 61: This needs to throw an error with `?` instead of using `unwrap()`.

Compiler is going to immediately complain, so you add this code: https://github.com/ouch-org/ouch/compare/main...bzip3#diff-97e25e2a0e41c578875856e97b659be2719a65227c104b992e3144efa000c35eR175-R187

And it will handle the error conversion.

The `BlockSize` should be unreachable if you're providing `mib(5)`, but just to be safe, please remove the `unreachable!` and add the `BlockSize` in the other treatment arm:

```diff
- Bz3Error::BlockSize => unreachable!(),
- Bz3Error::ProcessBlock(_) | Bz3Error::InvalidSignature => {
+ Bz3Error::BlockSize | Bz3Error::ProcessBlock(_) | Bz3Error::InvalidSignature => {
      FinalError::with_title("bzip3 error").detail(err.to_string()).into()
  }

**Review by marcospb19 — CHANGES_REQUESTED** [review]
Your code looks better than mine!

It's just missing the test code ([it's basically one line](https://github.com/ouch-org/ouch/compare/main...bzip3#diff-eb771f54a1fdc1c330a82981d7d27d1202a5cbe3a59581ff67af5e3361f5f90eR37))

And another requested change,

File: src/commands/compress.rs [hunk 2]
```diff
@@ -55,6 +55,10 @@ pub fn compress_files(
                 encoder,
                 level.map_or_else(Default::default, |l| bzip2::Compression::new((l as u32).clamp(1, 9))),
             )),
+            Bzip3 => Box::new(
+                // Unwrap is safe when using a valid block size
+                bzip3::write::Bz3Encoder::new(encoder, bytesize::ByteSize::mib(5).0 as usize).unwrap(),
```
marcospb19, line 60: 5 MiB sounds OK for now, although I'd prefer something around 10MB-20MB.

Right now we don't allow the user to provide compression levels for each format separately, something like `--levels bz=3,zst=7`, should we allow `.bz3` to accept a MB amount from the `--level` flag? It might be confusing with how `--level` is used by other formats.

@figsoda any thoughts?

figsoda: > should we allow `.bz3` to accept a MB amount from the `--level` flag? It might be confusing with how `--level` is used by other formats.

sounds good to me, this should be consistent with the `--block` flag of `bzip3`, assuming you mean `MiB`

alternatively we can just not respect the --level flag at all

figsoda, line 60: I think we should default to 16MiB here, which is consistent with bzip3's cli

also I'm not sure about using `bytesize` here

```suggestion
                bzip3::write::Bz3Encoder::new(encoder, 16 * 2.pow(20)).unwrap(),
```

freijon: done in [2289def](https://github.com/ouch-org/ouch/pull/522/commits/2289def77b29b6da11e277763826c82bbd41f7d3)

src/commands/compress.rs
freijon: Thank you for the feedback. I was not aware of your branch, looks like I could have saved some time ;) Anyway, I implemented your changes in [2289def](https://github.com/ouch-org/ouch/pull/522/commits/2289def77b29b6da11e277763826c82bbd41f7d3). Since they merged your PR over at [bzip3-rs](https://github.com/bczhc/bzip3-rs) I used this commit instead, at least until they publish a new crate version.

marcospb19: > looks like I could have saved some time ;)

Yep xD, I hope you enjoyed doing the coding tho.

I'll keep in mind to always create an issue or a draft PR to signal this to other people.

marcospb19:
Now the [CI is failing](https://github.com/ouch-org/ouch/actions/runs/6256978116/job/17010874173?pr=522) to build `bzip3-rs` because it doesn't have `bzip3` installed.

I don't know how to solve this for Ubuntu, MacOS and Windows, do we need to clone the repo and build it from scratch? :sob: 

Perhaps installing `nix` as a step and using [our nix-shell](https://github.com/ouch-org/ouch/blob/main/benchmarks/shell.nix)? cc @figsoda again.

figsoda:
Not sure what's the best way to do it on windows, It's available for brew so macos should be easy to take care of, it is only available on ubuntu 23.04+ so we might have to use brew for linux as well

Cargo.toml
```diff
@@ -14,6 +14,7 @@ description = "A command-line utility for easily compressing and decompressing f
 atty = "0.2.14"
 bstr = { version = "1.6.2", default-features = false, features = ["std"] }
 bzip2 = "0.4.4"
+bzip3 = { git = "https://github.com/bczhc/bzip3-rs", commit = "fa98dfce62d2271d0d3c91fd3592a912314a0df9" }
```
vrmiguel, line 17: Any chance we could use

```
bzip3 = { version = "0.8.1", features =  ["bundled"] }
```

Maybe that would help in the CI woes

figsoda: good catch! that will probably work

though I would recommend adding `bzip3/bundled` as a default feature instead of making it a hard dependency

marcospb19: Neat! I'm trying using this one.

@freijon I hope you don't mind, but I'm directly using your branch while trying to solve CI problems.

marcospb19 (commented):
Alright, CI problems are blocker number 4 (I'm trying to solve it).

But I just arrived at blocker number 5: tests are failing locally.

`cargo test --release`:
```
ERROR: corrupt xz stream
```

Cargo.toml
```diff
@@ -14,6 +14,7 @@ description = "A command-line utility for easily compressing and decompressing f
 atty = "0.2.14"
 bstr = { version = "1.6.2", default-features = false, features = ["std"] }
 bzip2 = "0.4.4"
+bzip3 = { git = "https://github.com/bczhc/bzip3-rs", commit = "fa98dfce62d2271d0d3c91fd3592a912314a0df9", features = ["bundled"] }
```
marcospb19, line 17: Alright, CI problems are blocker number 4 (I'm trying to solve it).

But I just arrived at blocker number 5: tests are failing locally.

`cargo test --release`:
```
ERROR: corrupt xz stream
```

Commenting in a piece of code to open a thread.

tests/integration.rs
```diff
@@ -34,6 +35,7 @@ enum DirectoryExtension {
 enum FileExtension {
     Bz,
     Bz2,
+    Bz3,
```
marcospb19, line 38: Alright, CI problems are blocker number 4 (I'm trying to solve them).

But I just arrived at blocker number 5: tests related to this line of code are failing.

`cargo test --release`:
```
ERROR: corrupt xz stream # file.zip.xz.bz3
ERROR: incomplete frame  # file.sz.zst.bz3
```

freijon: Is there anything I can help with?

marcospb19: yes, we need to figure out why `cargo test` is failing locally.

I believe this is an error in the `bzip3-rs` crate or `bzip3` itself, but we need enough info (or a minimal reproduction example) to open an issue in that repo.

A good point to start debugging is the `cargo test` output.

bczhc: Hi. I found this line fails in test `single_file`:
https://github.com/ouch-org/ouch/blob/6c2e1b6d820f4d898494e609555f0498592fd677/tests/integration.rs#L121

I took the `archive` file and did something like: `cat "/tmp/.tmpWMpRmq/file.sz.sz.zst.bz3.gz.bz2.bz.xz" | xz -d | bzip2 -d | bzip2 -d | gzip -d | bzip3 -d | zstd -d`, with no error occurred. But with `ouch` it says `ERROR: incomplete frame`; don't know why.

The Snappy stream (in this file it's `file.sz.sz`) is empty. Is empty stream considered valid for a Snappy format? Seems Snappy doesn't have an official CLI tool and I can't do decompression in a terminal.

BTW I noticed:
> Extension {
>         compression_formats: [
>             Lzma,
>         ],
>         display_text: "xz",
>     },

Actually `xz` has some difference with `LZMA` format (`xz --format=xz` vs `xz --format=lzma`). IMO `xz` is a container format of `LZMA2` format, which is a successor of the lagacy `LZMA` format.

marcospb19: Oh, wait!

Your fix to https://github.com/bczhc/bzip3-rs/pull/8 actually solved the `file.zip.xz.bz3` problem.

Now it's just the `.sz` one, I'm trying to debug it.

marcospb19:
I force-pushed a rebase to solve conflicts, and temporarily disabled some CI targets so that we can focus on the code problem for now.

tests/integration.rs
freijon: I ran `cargo test` with the latest commit on my machine and it runs fine, all tests pass. Is the issue only on CI?

marcospb19: It's failing locally for me, can you run `git log` and tell me at what commit you're at?

freijon: I downloaded a zip of `65d1229b6c13aeef9a4c16d73df1212fc8cbfc66` (freijon repo) which tested fine. But when I pull and rebase my repo it fails.

bczhc: OK guys. I also did a test and couldn't see any test failure now on my machine. Both at freijon's [`65d1229b6c13aeef9a4c16d73df1212fc8cbfc66`](https://github.com/freijon/ouch/commit/65d1229b6c13aeef9a4c16d73df1212fc8cbfc66) and `ouch-org:main` applied with freijon's changes (Console log: https://gist.github.com/bczhc/e4b356a2f05cae370e016c1fc47bcf73).

Can anyone tell me what's the issue we're now meeting?

bczhc: Sorry ignore what I said just now. After running `rustup update stable`, I can't reproduce the success case. And also sadly I forgot to check the rust toolchain version before `rustup update stable` so I don't know the toolchain version with which I just pasted the test. Searching in my shell history, the last time I ran `rustup update stable` is 2023-06-13, maybe the previous toolchain version is `stable-2023-06-01`. But with that version, it also fails in `single_file`. I don't know why, and how the hell I just passed those tests (as shown in the console log).

bczhc: I reproduced it. This is an unstable bug. Check it out: https://gist.github.com/bczhc/54a0ffbacd1de7591df40aaf92fc3287. Try searching `single_file ... ok` and `single_file ... failed`.

You can see the first two runs passed, but further runs failed.

Toolchain version: `stable-2023-06-01`

freijon: So is the root cause coming from [bczhc/bzip3-rs](https://github.com/bczhc/bzip3-rs) then?

bczhc: Yea very likely, my bad... The error is still from chained stuff and I'm still investing it. I tracked this in https://github.com/bczhc/bzip3-rs/issues/9.

bczhc: @freijon Hi. I've located the root cause and fixed it.

One interesting case:
```console
~/t/2 ❯ touch file                                                                                                           17:54:27
~/t/2 ❯ ouch c file file.sz.xz.bz3                                                                                           17:54:29
[INFO] Successfully compressed 'file.sz.xz.bz3'.
~/t/2 ❯ ouch d file.sz.xz.bz3 --yes && cat file                                                                              17:54:31
[INFO] Failed to confirm the format of `file` by sniffing the contents, file might be misnamed
[ERROR] premature eof
~/t/2 ❯                                                                                                                      17:54:35
~/t/2 ❯ rm *                                                                                                                 17:54:35
zsh: sure you want to delete all 2 files in /home/bczhc/t/2 [yn]? y
~/t/2 ❯ printf a > file                                                                                                      17:54:37
~/t/2 ❯ ouch c file file.sz.xz.bz3                                                                                           17:54:41
[INFO] Successfully compressed 'file.sz.xz.bz3'.
~/t/2 ❯ ouch d file.sz.xz.bz3 --yes && cat file                                                                              17:54:42
[INFO] Failed to confirm the format of `file` by sniffing the contents, file might be misnamed
[INFO] Successfully decompressed archive in current directory.
[INFO] Files unpacked: 1
a%
```

Only when the source file is empty the error encounters. I guess, when the source is empty, `XzEncoder` will call `write` once on `Bz3Encoder` with an empty buffer, so that `Bz3Encoder` produces an empty block. `Bz3Decoder` didn't handle empty blocks well, and finally leads to such a bug.

I now suppose empty blocks are totally valid in bzip3 format, and handle them. Another possible solution is reject `write` calls with zero-length buffer so that empty blocks won't be made.

marcospb19: Thanks for your help @bczhc :heart:, with all tests passing locally here, I'll mark this thread as resolved.

.github/workflows/build-and-test.yml
```diff
@@ -17,38 +17,38 @@ jobs:
     strategy:
       matrix:
         include:
-          - target: aarch64-unknown-linux-gnu
-            os: ubuntu-latest
-            no-zstd-thin: true
+          # - target: aarch64-unknown-linux-gnu
+          #   os: ubuntu-latest
+          #   no-zstd-thin: true
```
marcospb19, line 22: CI is passing because I temporarily disabled some targets that were giving build errors.

We need to re-enable these targets, otherwise, we'll end up dropping support for them (should we?).

Someone needs to go and uncomment these targets at`.github/workflows/build-and-test.yml`, and figure out how to make them build successfully, @freijon can you try doing that?

marcospb19 (changes requested):
Almost done, there is just one last problem:

.github/workflows/build-and-test.yml
freijon: I don't think I have the permission to do this?

marcospb19: I tried solving this one by changing the package version that was being installed by `apt`, no success tho.

marcospb19 (approved):
Yaay!

Thanks to @ttys3 the CI problem was solved in https://github.com/ouch-org/ouch/pull/646/commits/9ff7309cac4fe9b9cdaefab7c50979ee89f3758f and that unblocked me to finish this PR.

Congrats everyone for the collective effort :) .

Issue #152, opened 2021-11-02
Support `.rar`


c02y:
https://crates.io/crates/rar

figsoda:
This crate seems to be unmaintained, is there other alternatives?

vrmiguel:
RAR is a proprietary format so we'd only be able to decompress those (`unrar` is open source) but that sets a weird precedent since all other supported formats can be compressed and decompressed through `ouch` alone

@marcospb19 @GabrielSimonetto @SpyrosRoum  thoughts?

SpyrosRoum:
Hm yeah that would be weird. We *could* spawn a command to handle compressing it but that would introduce a runtime dependency and would also be different from the rest.
If we do that it might be worth locking it behind a feature flag as an extra warning

marcospb19:
I don't think we should support `.rar` because, unfortunately, we can't compress and decompress it reliably yet, and I don't think we will any time soon.

About spawning commands for external binaries, that may add a surprising amount of complexity for shipping, implementation and testing, making it even harder for people to contribute to our tool.

Also, it's possible that we would need to parse the STDERR output of the command to fit messages to our error treatment system, I wouldn't be surprised if this caused bugs.

If we cannot guarantee the quality of a feature, maybe the best decision is to skip it.

That's just my POV tho, if we get a nice crate for this and everybody agrees, we can add it.

sigmaSd:
Personally I think we should support tar and external binaries in general, because the way I see it is ouch wants to:
- be simple to use
- be the only compression/decompression tool that the user needs

If for each format that we don't support the user falls-back to another tool , it kind of defeat the second point

Also rar is not only the only common format we're missing, lzip for example is missing as well and there might be others

I think we can make a nice wrapper around all these tools that's clearly separated in its own module/crate and that exposes good api for ouch

Also when a native implementation arises we can just swap that

sigmaSd:
Hmm archlinux doesn't seem to even have rar (only unrar), I guess I don't feel strongly about this now

marcospb19:
> be the only compression/decompression tool that the user needs

That's not exactly true, we knew from the start that we wouldn't be able to cover all compression formats because there are a lot of them out there, but we already support the most important ones, for example, to my personal use cases, `ouch` covers 100% of the formats I need (`.tar.{zst, gz, xz}` and `.zip`).

I'd also say we fulfill the needs of at least 80% of our users with the formats we currently support.

[...]

PR #529, opened 2023-09-26
Add partial (listing and decompressing only) support for RAR

lmkra:
Attempt to address #152 using https://github.com/muja/unrar.rs crate.


CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# Changelog
+.# Changelog
```
marcospb19, line 1: Thanks for the PR!

This change looks unintentional.

CI is failing to compile the `unrar-sys` on `unknown-linux-musl`, does it require installing dependencies with `apt`?

lmkra: Ooops, that definitely was unintentional.

As for `unrar-sys` on musl-based systems, [it may be non-trivial](https://github.com/clux/muslrust/issues/114), but I'll look into it.

lmkra: I've tried to reproduce the build error, but failed to achieve that - linking statically against musl works fine for me.
(disclaimer: I do not use nor have at hand an Ubuntu machine)

I suspect it may be related to the specific setup of github actions environment.
I've pushed a workaround that links system `g++` to `musl-g++` to see whether any further errors will arise during tests.

lmkra: Thank you for running CI again, it helped me to reproduce the error.

Main issue here is that:
1) musl-tools does not provide C++ runtime linked against musl
2) libstdc++ included in github runner environment is linked against glibc

Easiest solution would be not to compile `-musl` builds directly on the runner system, but use cross environment (which includes full C++ toolchain linked against musl).

lmkra:
ARMv7 CI failure seems to be an upstream issue: https://github.com/muja/unrar.rs/issues/35

marcospb19:
Thanks for the follow ups, I temporarly disabled ARMv7 to see if CI succeeds in all other targets.

marcospb19:
Thanks for the follow ups!

I temporarly disabled ARMv7 to see if CI succeeds in other targets, it's failing on Windows, let me know whether you know how to fix it.

lmkra:
MSVC target on Windows should work fine, it's only MinGW one that fails.
I've proposed a fix for that: https://github.com/muja/unrar.rs/pull/37

In the meantime I've temporarily switched used unrar.rs version to my own fork, which contains all of upstream-proposed fixes.

lmkra:
Situation now seems to be that:
* x86_64-pc-windows-gnu target builds correctly
* ...but integrations tests fail due to ouch binary being dynamically linked to libstdc++-6.dll, which is provided by MinGW compiler and is not found in standard Windows shell search path.

While etting up $PATH variable to MinGW install directory or shipping aforementioned library together with ouch.exe could be considered as a workarounds, but I think that better solution would be to statically link against libstdc++ in ouch executable.
Just need to find out how to convince cargo to do it...

lmkra:
Seems to be working now 😃

marcospb19 (approved):
> Seems to be working now 😃

Amazing!

I think it's better not to merge right now because it would block new `Ouch` releases from happening, I can only publish a version on `crates.io` when all dependencies are fetched from `crates.io` itself, I guess we'll have to wait for a new `unrar` release (but we don't know when that comes out).

Does that make sense to you? :thinking:

lmkra:
Thanks!

And yes, that makes perfect sense to me 👍

lmkra:
New unrar has finally landed on crates.io :)
Therefore I've rebased this PR against current master branch and removed all temporary patching.

marcospb19:
That's awesome! Thank you for keeping an eye on that.

Cargo.toml
```diff
@@ -12,11 +12,11 @@ description = "A command-line utility for easily compressing and decompressing f
...
 [dependencies]
 atty = "0.2.14"
-bstr = { version = "1.8.0", default-features = false, features = ["std"] }
+bstr = { version = "1.6.2", default-features = false, features = ["std"] }
 bzip2 = "0.4.4"
-clap = { version = "4.4.8", features = ["derive", "env"] }
+clap = { version = "4.4.5", features = ["derive", "env"] }
 filetime = "0.2.22"
-flate2 = { version = "1.0.28", default-features = false }
+flate2 = { version = "1.0.27", default-features = false }
```
marcospb19, line 19: I did a last re-review, and found this which is unusual, I'm assuming it happened when you solved some merge conflict.

Before we merge, can you revert these?

marcospb19 (changes requested):

Cargo.toml
lmkra: Oops, must have slipped unnoticed during the rebase.
Should be ok now.

PR #556, opened 2023-11-26
Fixes CI artifacts and the manual-release workflow

Flat:

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -91,18 +91,14 @@ jobs:
           OUCH_ARTIFACTS_FOLDER: artifacts
           RUSTFLAGS: -C strip=symbols
 
-      - name: Upload binary
+      - name: Upload artifacts
         uses: actions/upload-artifact@v3
         with:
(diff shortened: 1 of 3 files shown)


marcospb19:
I tried using this but it failed, here's [the run result](https://github.com/ouch-org/ouch/actions/runs/7050257974/job/19190539504):

```sh
Run scripts/package-release-assets.sh
  scripts/package-release-assets.sh
  shell: /usr/bin/bash -e {0}
cp: cannot stat 'ouch-aarch64-pc-windows-msvc.exe/artifacts': No such file or directory
Error: Process completed with exit code 1.
```

Flat:
Did you run it based off a build-and-test from this PR? If you use `6998396623` it should work fine. This is due to the changes made to the build-and-test to actually include the manuals and completions as artifacts.

marcospb19 (approved):
You're such a beast!

Thank you so much for this, and for explaining to me why my test was failing :sweat_smile:.

Flat:
Glad I could help! Hopefully it makes it easier for releases on limited time, haha.

PR #573, opened 2023-12-05
update README

### LINKED ISSUE #568: The "Dependencies" section of the README seems outdated.
https://github.com/ouch-org/ouch/blob/0dbbd3b8825d1bb8f985b3e384cdd987d5f4d40f/README.md?plain=1#L166-L176

This mentions a download script that no longer seems to exist. Also, it's unclear what systems these dependencies are necessary on (what about each version of the official binaries?), or whether these are build-time or runtime dependencies.

marcospb19:
fix dependencies section and make other minor changes

fixes #568.


ilyagr:
Thank you! I spotted one sentence I would word slightly differently. It's only a very minor nit, but I will send a PR in a second.

Issue #576, opened 2023-12-07
Support for encrypted archives


TLATER:
Some archive types like 7z can be encrypted - looks like `sevenz-rust` [at least partially supports en-/decrypting such archives](https://crates.io/crates/sevenz-rust#decompress-a-encrypted-7z-file).

I don't personally have much of a use case for encrypting such archives, but this has really become a swiss army knife for decompressing various archives - it's annoying to need to install other packages when I stumble upon an encrypted one, so it would be really nice if at least decryption was supported.

marcospb19:
related to #326

PR #646, opened 2024-03-24
feature: support decompress and list archive file with password

ttys3:
resolve https://github.com/ouch-org/ouch/issues/576

- [x] https://github.com/muja/unrar.rs/pull/54

~~there's a bug in unrar.rs I submitted a PR https://github.com/muja/unrar.rs/pull/49~~

~~we need to wait upstream get merged so that we can have CJK filename support for rar files~~

### DIFF
--- a/Cross.toml
+++ b/Cross.toml
@@ -1,2 +1,9 @@
 [build.env]
 passthrough = ["RUSTFLAGS", "OUCH_ARTIFACTS_FOLDER"]
+
+[target.aarch64-unknown-linux-gnu]
+# Translates to `ghcr.io/cross-rs/aarch64-unknown-linux-gnu:edge`
+image = "ghcr.io/cross-rs/aarch64-unknown-linux-gnu:edge"
+
+[target.armv7-unknown-linux-gnueabihf]
+image = "ghcr.io/cross-rs/armv7-unknown-linux-gnueabihf:edge"
--- a/Cross.toml
+++ b/Cross.toml
@@ -2,7 +2,6 @@
 passthrough = ["RUSTFLAGS", "OUCH_ARTIFACTS_FOLDER"]
 
 [target.aarch64-unknown-linux-gnu]
-# Translates to `ghcr.io/cross-rs/aarch64-unknown-linux-gnu:edge`
 image = "ghcr.io/cross-rs/aarch64-unknown-linux-gnu:edge"
 
 [target.armv7-unknown-linux-gnueabihf]
--- a/src/archive/rar.rs
+++ b/src/archive/rar.rs
@@ -61,7 +61,7 @@ pub fn list_archive(
 
             Ok(FileInArchive { path, is_dir })
         }).collect::<Vec<_>>().into_iter(),
-        Err(e) => vec![Err(Error::UnsupportedFormat {reason:e.to_string()})].into_iter(),
+        Err(e) => vec![Err(Error::UnrarError{reason: e.to_string()})].into_iter(),
     }
 }
 
--- a/src/archive/rar.rs
+++ b/src/archive/rar.rs
@@ -55,14 +55,17 @@ pub fn list_archive(
     };
 
     let result = match archive.open_for_listing() {
-        Ok(iter) => iter.map(|item| {
-            let item = item?;
-            let is_dir = item.is_directory();
-            let path = item.filename;
+        Ok(iter) => iter
+            .map(|item| {
+                let item = item?;
+                let is_dir = item.is_directory();
+                let path = item.filename;
 
-            Ok(FileInArchive { path, is_dir })
-        }).collect::<Vec<_>>().into_iter(),
-        Err(e) => return Err(Error::UnrarError{reason: e.to_string()}),
+                Ok(FileInArchive { path, is_dir })
+            })
+            .collect::<Vec<_>>()
+            .into_iter(),
+        Err(e) => return Err(Error::UnrarError { reason: e.to_string() }),
     };
 
     Ok(result)
--- a/src/archive/rar.rs
+++ b/src/archive/rar.rs
@@ -54,21 +54,15 @@ pub fn list_archive(
         None => Archive::new(archive_path),
     };
 
-    let result = match archive.open_for_listing() {
-        Ok(iter) => iter
-            .map(|item| {
-                let item = item?;
-                let is_dir = item.is_directory();
-                let path = item.filename;
+    let archive = archive.open_for_listing()?;
 
-                Ok(FileInArchive { path, is_dir })
-            })
-            .collect::<Vec<_>>()
-            .into_iter(),
-        Err(e) => return Err(Error::UnrarError { reason: e.to_string() }),
-    };
+    Ok(archive.map(|item| {
+        let item = item?;
+        let is_dir = item.is_directory();
+        let path = item.filename;
 
-    Ok(result)
+        Ok(FileInArchive { path, is_dir })
+    }))
 }
 
 pub fn no_compression() -> Error {
--- a/src/archive/rar.rs
+++ b/src/archive/rar.rs
@@ -53,13 +53,16 @@ pub fn list_archive(
         Some(password) => Archive::with_password(archive_path, password),
         None => Archive::new(archive_path),
     };
-    archive.open_for_listing().expect("cannot open archive").map(|item| {
-        let item = item?;
-        let is_dir = item.is_directory();
-        let path = item.filename;
+    match archive.open_for_listing() {
+        Ok(iter) => iter.map(|item| {
+            let item = item?;
+            let is_dir = item.is_directory();
+            let path = item.filename;
 
-        Ok(FileInArchive { path, is_dir })
-    })
+            Ok(FileInArchive { path, is_dir })
+        }).collect::<Vec<_>>().into_iter(),
+        Err(e) => vec![Err(Error::UnsupportedFormat {reason:e.to_string()})].into_iter(),
+    }
 }
 
 pub fn no_compression() -> Error {
--- a/src/archive/rar.rs
+++ b/src/archive/rar.rs
@@ -48,12 +48,13 @@ pub fn unpack_archive(
 pub fn list_archive(
     archive_path: &Path,
     password: Option<&[u8]>,
-) -> impl Iterator<Item = crate::Result<FileInArchive>> {
+) -> crate::Result<impl Iterator<Item = crate::Result<FileInArchive>>> {
     let archive = match password {
         Some(password) => Archive::with_password(archive_path, password),
         None => Archive::new(archive_path),
     };
-    match archive.open_for_listing() {
+
+    let result = match archive.open_for_listing() {
         Ok(iter) => iter.map(|item| {
             let item = item?;
             let is_dir = item.is_directory();
@@ -61,8 +62,10 @@ pub fn list_archive(
 
             Ok(FileInArchive { path, is_dir })
         }).collect::<Vec<_>>().into_iter(),
-        Err(e) => vec![Err(Error::UnrarError{reason: e.to_string()})].into_iter(),
-    }
+        Err(e) => return Err(Error::UnrarError{reason: e.to_string()}),
+    };
+
+    Ok(result)
 }
 
 pub fn no_compression() -> Error {
--- a/src/archive/rar.rs
+++ b/src/archive/rar.rs
@@ -8,7 +8,12 @@ use crate::{error::Error, list::FileInArchive, utils::logger::info};
 
 /// Unpacks the archive given by `archive_path` into the folder given by `output_folder`.
 /// Assumes that output_folder is empty
-pub fn unpack_archive(archive_path: &Path, output_folder: &Path, password: Option<impl AsRef<[u8]>>, quiet: bool) -> crate::Result<usize> {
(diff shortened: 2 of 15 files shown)


ttys3:
seems the upsteam no response currently. can we use a fork of unrar.rs for temp ?

ttys3:
finally get tests issue on musl libc, aarch64, armv7 and macOS resolved.

ttys3:
good news, After waiting for a rather long time, the upstream finally merged my PR. 

we can continue this PR now

ttys3:
all check is green now, cheers!

I updated cross-rs image for   `aarch64-unknown-linux-gnu` and `armv7-unknown-linux-gnueabihf` because its gcc version is too old.

see 
https://github.com/cross-rs/cross/blob/4645d937bdae6952d9df38eff3ecb91fd719c3bd/README.md?plain=1#L323

image

Cross.toml
```diff
@@ -1,2 +1,9 @@
 [build.env]
 passthrough = ["RUSTFLAGS", "OUCH_ARTIFACTS_FOLDER"]
+
+[target.aarch64-unknown-linux-gnu]
+# Translates to `ghcr.io/cross-rs/aarch64-unknown-linux-gnu:edge`
```
marcospb19, line 5: This comment is redundant

src/cli/args.rs
```diff
@@ -41,6 +41,10 @@ pub struct CliArgs {
     #[arg(short, long, global = true)]
     pub format: Option<OsString>,
 
+    /// decompress or list with password
+    #[arg(short = 'p', long = "password", global = true)]
+    pub password: Option<String>,
```
marcospb19, line 46: Ideally, here we'd use OsString to support non-UTF-8 passwords (that's why zip and rar ask for `&[u8]` and not `&str`)

Idk where Clap documents that it supports OsString but it does :smile:.

(After the change you can use `args.password.as_ref().map(|string| string.as_bytes()),` in the two places it will complain.)

---

This suggestion goes together with the other suggestion of using `Option<&[u8]>` in list and decompression function signatures instead of `impl AsRef<[u8]>`.

src/archive/rar.rs
```diff
@@ -8,10 +8,22 @@ use crate::{error::Error, list::FileInArchive, utils::logger::info};
 
 /// Unpacks the archive given by `archive_path` into the folder given by `output_folder`.
 /// Assumes that output_folder is empty
-pub fn unpack_archive(archive_path: &Path, output_folder: &Path, quiet: bool) -> crate::Result<usize> {
+pub fn unpack_archive(
+    archive_path: &Path,
+    output_folder: &Path,
+    password: Option<impl AsRef<[u8]>>,
```
marcospb19, line 14: We are only passing passwords with a single type, so we don't need this to be generic, you can provide `Option<&[u8]>` here directly.

src/archive/rar.rs
```diff
@@ -35,17 +47,22 @@ pub fn unpack_archive(archive_path: &Path, output_folder: &Path, quiet: bool) ->
...
-        .map(|item| {
-            let item = item?;
-            let is_dir = item.is_directory();
-            let path = item.filename;
-
-            Ok(FileInArchive { path, is_dir })
-        })
+pub fn list_archive(
+    archive_path: &Path,
+    password: Option<impl AsRef<[u8]>>,
```
marcospb19, line 52: Same here and on the rest of decompression and listing functions.

src/archive/rar.rs
```diff
@@ -35,17 +47,22 @@ pub fn unpack_archive(archive_path: &Path, output_folder: &Path, quiet: bool) ->
...
+) -> impl Iterator<Item = crate::Result<FileInArchive>> {
+    let password = password.as_ref().map(|p| p.as_ref());
```
marcospb19, line 54: I believe you can remove 4 places this mapping appears after replacing `Option<impl AsRef<[u8]>>` by `Option<&[u8]>`

src/archive/sevenz.rs
```diff
@@ -150,7 +159,51 @@ where
...
-    })?;
+    };
+
+    let password = password.as_ref().map(|p| p.as_ref());
+
+    match password {
+        Some(password) => sevenz_rust::decompress_with_extract_fn_and_password(
+            reader,
+            output_path,
+            sevenz_rust::Password::from(password.to_str().unwrap()),
```
marcospb19, line 170: inside of the `Some` branch of this `match`, we should try converting the provided bytes to UTF-8 or throw a small `FinalError` with the brief description that `sevenz_rust` expects utf-8 valid passwords.

you can do this with a `let-else` construct with a return inside

src/archive/rar.rs
```diff
@@ -35,17 +47,22 @@ pub fn unpack_archive(archive_path: &Path, output_folder: &Path, quiet: bool) ->
...
+    let archive = match password {
+        Some(password) => Archive::with_password(archive_path, password),
+        None => Archive::new(archive_path),
+    };
+    archive.open_for_listing().expect("cannot open archive").map(|item| {
```
marcospb19, line 59: This was already present, but please replace this `expect` by a `?`

src/archive/sevenz.rs
```diff
@@ -150,7 +159,51 @@ where
...
 
     Ok(count)
 }
+
+/// List contents of `archive_path`, returning a vector of archive entries
+pub fn list_archive(
+    archive_path: &Path,
+    password: Option<impl AsRef<[u8]>>,
+) -> impl Iterator<Item = crate::Result<FileInArchive>> {
+    let reader = fs::File::open(archive_path).unwrap();
```
marcospb19, line 184: There are four `.unwrap()`s in this function that can't exist, so this function should return `Result` and the two should be raised with the try operator (`"?"`).

src/archive/sevenz.rs
```diff
@@ -150,7 +159,51 @@ where
...
+            is_dir: entry.is_directory(),
+        }));
+        Ok(true)
+    };
+
+    match password {
+        Some(password) => sevenz_rust::decompress_with_extract_fn_and_password(
+            reader,
+            ".",
+            sevenz_rust::Password::from(password.to_str().unwrap()),
```
marcospb19, line 201: Same case for utf-8 invalid. reuse code from that previous let-else (is there a better way?)

src/archive/sevenz.rs
```diff
@@ -150,7 +159,51 @@ where
...
+            entry_extract_fn,
+        )
+        .unwrap(),
+        None => sevenz_rust::decompress_with_extract_fn(reader, ".", entry_extract_fn).unwrap(),
```
marcospb19, line 205: The other 2 unwraps

src/archive/zip.rs
```diff
@@ -28,16 +29,29 @@ use crate::{
...
 
+    let password = password.as_ref().map(|p| p.as_ref().to_owned());
+
     for idx in 0..archive.len() {
-        let mut file = archive.by_index(idx)?;
+        let mut file = match password.clone() {
+            Some(password) => archive
+                .by_index_decrypt(idx, password.to_owned().as_bytes())
+                .unwrap()
+                .map_err(|_| zip::result::ZipError::UnsupportedArchive("Password required to decrypt file"))?,
```
marcospb19, line 52: You can replace these 2 lines by the double try operator `??`.

No need to map the error because `zip` will give the detailed information of why this failed, and our error treatment is set in a way that gives a good error message.

src/archive/zip.rs
```diff
@@ -105,11 +122,21 @@ where
...
     let (tx, rx) = mpsc::channel();
     thread::spawn(move || {
         for idx in 0..archive.len() {
             let maybe_file_in_archive = (|| {
-                let file = match archive.by_index(idx) {
+                let zip_result = match password.clone() {
+                    Some(password) => archive
+                        .by_index_decrypt(idx, password.to_owned().clone().as_bytes())
+                        .unwrap()
+                        .map_err(|_| zip::result::ZipError::UnsupportedArchive("Password required to decrypt file")),
```
marcospb19, line 135: Ditto.

src/archive/zip.rs
```diff
@@ -28,16 +29,29 @@ use crate::{
...
 
     let mut unpacked_files = 0;
 
+    let password = password.as_ref().map(|p| p.as_ref().to_owned());
+
     for idx in 0..archive.len() {
-        let mut file = archive.by_index(idx)?;
+        let mut file = match password.clone() {
+            Some(password) => archive
+                .by_index_decrypt(idx, password.to_owned().as_bytes())
```
marcospb19, line 50: This to_owned shouldn't be necessary considering we have another 5 lines above.

marcospb19 (changes requested):
Awesome PR! It covers a lot of things right and I appreciate the effort.

> I updated cross-rs image for aarch64-unknown-linux-gnu and armv7-unknown-linux-gnueabihf because its gcc version is too old.

That's great!! I had no idea what was going on and this was blocking another PRs, that's a huge win, ty ty.

---

Although the PR is great, there is some rough edges to adjust.

I see you've done a hell lot for me in this one, so let me know if you want me to make the final adjustments for you :wink:.

src/archive/zip.rs
```diff
@@ -105,11 +119,20 @@ where
...
+
     let (tx, rx) = mpsc::channel();
     thread::spawn(move || {
         for idx in 0..archive.len() {
             let maybe_file_in_archive = (|| {
-                let file = match archive.by_index(idx) {
+                let zip_result = match password.clone() {
+                    Some(password) => archive
+                        .by_index_decrypt(idx, password.as_bytes()).ok()?
+                        .map_err(|_| zip::result::ZipError::UnsupportedArchive("Password required to decrypt file")),
```
marcospb19, line 131: Ah, now I see the problem, this closure is returning `Option<Result<_>>` instead of `Result<Option<_>>`, that's why you couldn't do `??` and just return the error.

Well, I guess there is no reason for us to here return `Option`, my bet is that `by_index` doesn't fail because `idx` is always in bounds.

Anyways, with `.ok()?` we do lose context on why this failed, we need `?` without the `.ok()?`, can you try making this closure return `Result<FileInArchive>` instead of `Option<Result<FileInArchive>>`?

src/archive/sevenz.rs
```diff
@@ -150,7 +160,59 @@ where
...
+    }
 
     Ok(count)
 }
+
+/// List contents of `archive_path`, returning a vector of archive entries
+pub fn list_archive(
+    archive_path: &Path,
+    password: Option<&[u8]>,
+) -> impl Iterator<Item = crate::Result<FileInArchive>> {
```
marcospb19, line 184: This should return `Result<impl Iterator<Item = Result<FileInArchive>>>` so you don't have to return a `vec![Err(Error::InvalidPassword(...))]`, which is a little hacky.

src/archive/sevenz.rs
ttys3: resolved in 5f11600

src/archive/rar.rs
ttys3: done

src/archive/sevenz.rs
ttys3: > This should return `Result<impl Iterator<Item = Result<FileInArchive>>>` so you don't have to return a `vec![Err(Error::InvalidPassword(...))]`, which is a little hacky.

yes, but if in this way, we need to change all the funcs return type, is this OK ?

src/archive/zip.rs
ttys3: yes, this is annoying.  
because:

```rs
    pub fn by_index_decrypt<'a>(
        &'a mut self,
        file_number: usize,
        password: &[u8],
    ) -> ZipResult<Result<ZipFile<'a>, InvalidPassword>> {
        self.by_index_with_optional_password(file_number, Some(password))
    }

    /// Get a contained file by index
    pub fn by_index(&mut self, file_number: usize) -> ZipResult<ZipFile<'_>> {
        Ok(self
            .by_index_with_optional_password(file_number, None)?
            .unwrap())
    }
```

but we can not call `by_index_with_optional_password` because it is not public exported.

ttys3: done in 164f800

src/archive/sevenz.rs
ttys3: changed for `sevenz` and `rar`,  do I need to change the return type for other types?

ttys3: > changed for `sevenz` and `rar`, do I need to change the return type for other types?
> 
> I mean, to keep they are the same

also done for `tar` and `zip`

src/archive/zip.rs
marcospb19: Oh I see what you mean now, yes that is annoying.

Code is good tho ty.

src/archive/sevenz.rs
marcospb19: That wasn't needed but it's OK, in a scenario like this, if a function doesn't fail, it's better to keep it returning the type directly, and make only the fallible functions return `Result`.

There are other cases where you might want to keep a generic interface fallible, but it doesn't apply here since we are not a library, etc.

marcospb19 (approved):
Thanks for the follow-ups and all the effort!

src/cli/args.rs
marcospb19: Argh this suggestion also seems wrong, my bad.

PR #654, opened 2024-04-01
CI refactor: use install-action & cargo-hack

cyqsimon:
## cargo-hack

I recently discovered [cargo-hack](https://github.com/taiki-e/cargo-hack/) which has the powerful (😉) feature of `--feature-powerset`, which allows easy testing of all possible combinations of features. This has two main benefits compared to what we are doing now:
1. much faster than the "one runner per feature set" approach, because it saves a lot of duplicate steps
2. more reliable than bespoke shell scripting

## install-action

[install-action](https://github.com/taiki-e/install-action), maintained by the same developer, is a convenient way to install tools like cross and cargo-hack.

## Problems

Currently cargo-hack does not seem to support keeping the compiled artifacts for each run, which means we will be losing some functionality in our CI. This is why this PR is still a draft. I have submitted [an issue](https://github.com/taiki-e/cargo-hack/issues/243) to ask about this.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -12,8 +12,6 @@ jobs:
   build:
     name: build
     runs-on: ${{ matrix.os }}
-    env:
-      CARGO: cargo
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


marcospb19:
Nice! You can ping me if there are any updates regarding https://github.com/taiki-e/cargo-hack/issues/243.

PR #690, opened 2024-07-14
Fix confusion between empty/nonempty test.

rcorre:
The single_empty_file test was writing random content to the file,
whereas the single_file test was writing an empty file.
Only the latter tested different levels, so I figured that test
was the one that should be using actual file content.


marcospb19:
Huh! :sweat_smile: that's odd! Thanks for the fix.

Issue #722, opened 2024-10-01
[RFC] Landlock sandbox support


valoq:
(De)Compression is historically one of the most common and easy to exploit attack vectors given that usually means to parse untrusted data. While around 70% of software vulnerabilities is related to memory management vulnerabilities, there are still plenty of issues related to logic bugs ([Examples](https://github.com/python/cpython/issues/94531)). Additionally, even the rust (de)compression libraries often make use of the unsafe keyword, leaving some potential for memory issues. 

To address these risks, one very simple to use solution for modern Linux system is [landlock](https://docs.rs/landlock/latest/landlock/) which allows a process to restrict its access to the filesystem. Implementing this into ouch would require little effort and provide a decent safety measure against potential issues that allow writing outside the intended directories.

(Note there is an equal solution for OpenBSD called [unveil](https://man.openbsd.org/unveil) and Windows has an app isolation feature that can be used to implement similar functionality in case this feature is wanted on other platforms as well.)

For a full sandbox that implements proper defenses against code execution vulnerabilities, additional measures like syscall filter would be required https://docs.rs/seccomp/latest/seccomp/ (Seccomp does require extensive testing on all architectures) 
This could be implemented as an additional step, though support for landlock would be the logical first step that addresses most risks with less efforts.

PR #723, opened 2024-10-01
DRAFT: Add landlock support

valoq:
This PR adds landlock filesystem isolation to ouch as discussed in #722 

At the moment this is just a quickly hacked implementation to demonstrate the use of landlock in ouch.
It restricts the entire filesystem to be read only and only permits write actions in the current working directory of the process.

In order to test the isolation feature, use the `-d `option to write the decompressed files to a path outside of `$PWD`
A final implementation would address the `-d` option as well to allow writing to specified output directories, but I left it untouched for now to allow an easy demonstration/test of the landlock feature.

Todo:
- restrict write permissions to the .tmp-XXXXXX directory, including permissions for said directory to rename itself to the final name LANDLOCK_ACCESS_FS_MAKE_DIR (This should avoid writing to any control files in $HOME like .bashrc)
- Identify all special cases that require permissions (see failed unit tests)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Categories Used:
 - Provide Nushell completions (packages still need to install them) [\#827](https://github.com/ouch-org/ouch/pull/827) ([FrancescElies](https://github.com/FrancescElies))
 - Support `.lz` decompression [\#838](https://github.com/ouch-org/ouch/pull/838) ([zzzsyyy](https://github.com/zzzsyyy))
 - Support `.lzma` decompression (and fix `.lzma` being a wrong alias for `.xz`) [\#838](https://github.com/ouch-org/ouch/pull/838) ([zzzsyyy](https://github.com/zzzsyyy))
+- Add landlock support for linux filesystem isolation [\#723](https://github.com/ouch-org/ouch/pull/723) ([valoq](https://github.com/valoq)) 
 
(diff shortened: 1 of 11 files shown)


valoq:
This feature should now work and is ready for some testing by those interested.

At the moment the flowing cases are supported:

- Listing files in an archive will have write restriction to "/tmp" since some formats create temporary files there
- Decompressing restricts write access to either the specified directory (-d option) or the current working directory where ouch is executed

TODO:

- Fix CI test
- implement landlock unit test
- Move tmpfile for rar (and possible other formats) to the global decompress (and list) space to provide to landlock

Future improvements:
The landlock restriction can potentially be more strict when no output directory is specified.
Currently, this results in write permissions for the current working directory, which is can be unsafe if ouch is run in $HOME for example. By allowing write actions only to the temporary directory while allowing LANDLOCK_ACCESS_FS_MAKE_DIR in the current working directory for later renaming of the temporary directory will fix this.
However at the moment ouch does not always use a temporary directory in all cases, which requires this to be changed first for effective isolation. Also see https://github.com/onekey-sec/unblob/pull/597

Additionally the isolation for the list command can be further restricted to only allow write permissions in the temporary directory created in "/tmp"

valoq:
The CI currently fails due to #852

valoq:
Closed in favor of #995

PR #745, opened 2024-11-18
fix: warnings not shown before stdin interaction

marcospb19:


marcospb19:
cc @AntoniosBarotsis this log wasn't being shown before the STDIN question:

https://github.com/ouch-org/ouch/blob/c18c6c8b0eb996ec76e84f8f2b9e97e56499531d/src/check.rs#L38-L45

that's something we missed

src/utils/logger.rs
```diff
@@ -202,6 +222,11 @@ mod logger_thread {
                     flush_logs_to_stderr(&mut buffer);
                     break;
                 }
+                Message::Flush { finished_barrier } => {
+                    flush_logs_to_stderr(&mut buffer);
+                    finished_barrier.wait();
+                    break;
```
marcospb19, line 228: oops this "break" shouldn't be here

https://github.com/ouch-org/ouch/pull/748

PR #752, opened 2024-11-18
refac: simplify logger shutdown system

marcospb19:


marcospb19:
cc @AntoniosBarotsis again just in case you're interested in this

PR #757, opened 2024-11-30
feat(cli): add option to remove source file after decompression

ttys3:
add cli flag to allow auto remove source file after decompression, this is often the common usage

like https://github.com/aonez/Keka on Mac, simple double click and file extracted, source file removed.

### DIFF
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -90,7 +90,7 @@ pub enum Subcommand {
         output_dir: Option<PathBuf>,
 
         /// Remove the source file after successful decompression
-        #[arg(short = 'r', long, default_value_t = false)]
+        #[arg(short = 'r', long)]
         remove: bool,
     },
     /// List contents of an archive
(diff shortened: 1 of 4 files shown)


src/cli/args.rs
```diff
@@ -88,6 +88,10 @@ pub enum Subcommand {
         /// Place results in a directory other than the current one
         #[arg(short = 'd', long = "dir", value_hint = ValueHint::FilePath)]
         output_dir: Option<PathBuf>,
+
+        /// Remove the source file after successful decompression
+        #[arg(short = 'r', long, default_value_t = false)]
```
marcospb19, line 93: false is the default value for booleans right? :thinking:

marcospb19 (approved):
great! can you add an entry to CHANGELOG.md for this feature?

src/cli/args.rs
ttys3: just want it to be more explicit. because remove file is a danger operation.

yes, I'm OK to remove that.

ttys3:
> great! can you add an entry to CHANGELOG.md for this feature?

done

marcospb19 (approved):
> done

Thank you!

PR #759, opened 2024-11-30
fix: fix decompress issue with rar files which include CJK filenames in it

ttys3:
update unrar crate to [fix decompress issue with rar files which include CJK filenames in it](https://github.com/muja/unrar.rs/commit/5e6b83d91982234d3686b235dcbc7989402a88fd)

prev we fixed the CJK filename issue for the archive filename itself

this time, it is the filename in the archive.

that's to day, if `foo.rar` contains `❤️.txt`  in it, the decompression will fail

we need wait unrar crate to release a new version


marcospb19:
Yeah, we need to wait for their release otherwise this can block a new version of Ouch.

ttys3:
OK, I changed the PR to WIP status

marcospb19 (approved):
looks like you also updated all deps, I guess it's a good thing tho

marcospb19:
ty, I see you created another PRs, sorry, I might take some time to take a proper look at them, busy week, but I'll catch up

PR #760, opened 2024-11-30
feat: add concurrent working threads option to CLI args

ttys3:
add concurrent working threads option to CLI args

### DIFF
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -46,7 +46,7 @@ pub struct CliArgs {
     pub password: Option<OsString>,
 
     /// cocurrent working threads
-    #[arg(short = 't', long, global = true)]
+    #[arg(short = 'c', long, global = true)]
     pub threads: Option<usize>,
 
     // Ouch and claps subcommands
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -45,6 +45,10 @@ pub struct CliArgs {
     #[arg(short = 'p', long = "password", global = true)]
     pub password: Option<OsString>,
 
+    /// cocurrent working threads
+    #[arg(short = 't', long, global = true)]
+    pub threads: Option<usize>,
+
     // Ouch and claps subcommands
     #[command(subcommand)]
     pub cmd: Subcommand,
@@ -138,6 +142,7 @@ mod tests {
             format: None,
             // This is usually replaced in assertion tests
             password: None,
+            threads: None,
             cmd: Subcommand::Decompress {
                 // Put a crazy value here so no test can assert it unintentionally
                 files: vec!["\x00\x11\x22".into()],
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -28,6 +28,13 @@ impl CliArgs {
 
         set_accessible(args.accessible);
 
+        if let Some(threads) = args.threads {
+            rayon::ThreadPoolBuilder::new()
+                .num_threads(threads)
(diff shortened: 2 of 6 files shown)


ttys3:
close due to `ThreadPoolBuildError { kind: GlobalPoolAlreadyInitialized }`

`spawn_logger_thread()` called and which uses rayon, the the global pool already initialized there

ttys3:
maybe, could use https://docs.rs/rayon/latest/rayon/struct.ThreadPoolBuilder.html#method.build_scoped ?

ttys3:
https://github.com/rayon-rs/rayon/blob/main/FAQ.md#how-many-threads-will-rayon-spawn

marcospb19:
You could call `parse_and_validate_args`, build the threadpool, and then call `spawn_logger_thread`, as `parse_and_validate_args` doesn't require the logger thread to be alive yet.

ttys3:
> You could call `parse_and_validate_args`, build the threadpool, and then call `spawn_logger_thread`, as `parse_and_validate_args` doesn't require the logger thread to be alive yet.

yes, that's a workaround.

but the problem is, I did not expect the `logger` and the `business code` share the **same** global rayon threadpool.

because, when I set the `threads` (the one I want to add) to 1, what I mean it decompress the files one by one, to reduce disk io impact.

but the logger may want more concurrent threads, I did not know this will cause problem for the logger. or, this is OK we do this?

ttys3:
maybe we could use async? that way, the async logger wouldn't have to depend on rayon to achieve that.

marcospb19:
> But since the logger may require more concurrent threads, I'm unsure if this will cause problems for it. Is it okay for us to do this?

Oh, I see your concern, I overlooked the case where parallelism is set to 1.

The solution is then to use `thread::spawn` instead of `rayon::spawn` in the logger thread, so we limit `rayon` threads just to the decompression.

Setting up an async logger wouldn't help because we'd need at least one thread in the async runtime threadpool, and that would be equivalent (but more complicated) than just having a `thread::spawn` thread separated for it.

ttys3:
Oops, seems that the UI tests failed

marcospb19:
image

it changed what the `--help` looks like and we have that on UI tests, so you need to run `cargo insta test` and `cargo insta review`, or update those snapshots manually till they match.

I know this is kinda annoying, but makes life easier in the long term.

ttys3:
UI tests fixed

marcospb19 (approved):
Ty, let me know if you think the default shouldn't be all cores.

ttys3:
> Ty, let me know if you think the default shouldn't be all cores.

If the compression algorithm itself is fast enough (and can effectively utilize all cores), I believe we should have all CPU cores focus on processing a single compressed document instead of dividing the CPU cores to handle multiple documents. This way, a single document can be decompressed more quickly. Meanwhile, in this scenario, if we decompress all documents concurrently, the bottleneck will not be the CPU, but rather disk IO, and threads can easily enter the D-state （the state in top or htop, mostly iowait）.

If the compression algorithm cannot efficiently utilize the CPU (for example, when decompressing with 7z and rar, despite claiming to support multithreading, they cannot fully utilize the CPU, and overall decompression speed is very slow). Then decompressing documents one by one will be relatively slow.

Therefore, I feel that both approaches (processing one task at a time vs. concurrently processing multiple compressed documents) do not have absolute pros and cons; it depends on the situation.

PR #764, opened 2025-02-12
fix: Use BufWriter for list output

### LINKED ISSUE #702: use buffered output in `list` command
https://github.com/ouch-org/ouch/blob/b2d0ff75c89a1dd2d60938d0a490b320efcc1a43/src/list.rs#L35

All output to this file should be done with `BufWriter`.

(On top of that, `writeln!` shouldn't be mixed with `print!`, we should only use the former.)

killercup:
Also replaces one `print` call with `write`.

Fixes #702


marcospb19 (approved):
Thanks!!!

PR #779, opened 2025-03-24
feat: Add rename option in overwrite menu

### LINKED ISSUE #767: file conflicts: add option to rename (add suffixes)
Instead of asking `overwrite? (y/n)`

Ask `overwrite? (y/n/rename)`

Which accepts 'r' as the answer.

Later we can add a flag `--rename` (can bikeshed) to skip this question (same or different PR).

### Proposed renaming behavior:

The file `path/to/file.ext` should be renamed to `path/to/file-1.ext`, but only for the file and folder paths that actually conflict (we can't detect conflicts in advance with how unpacking works right now to ensure it's applied to all files).

To clarify: if you add the suffix to `folder-1`, you don't need to add suffixes to files inside of that folder.

If `file-1.ext` already exists, bump the number and try that again, this might require multiple tries, try up to `99` and after just crash with a decent error message saying that we hit the limit and something strange is probably happening, people shouldn't rely on this behavior cause it's too wasteful and leaves tons of uncompressed archives taking space.

Prior discussion: #766.

talis-fb:
Closes #767 

## Examples

> Setup
> ```sh
> echo '...' > something.txt
> ouch compress something.txt something.zip
> ```

* Basic example:
image

* Accessible mode:
image

* Auto increment the final number when there are another conflict in renamed name
image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,6 +73,7 @@ test-strategy = "0.4.0"
 default = ["use_zlib", "use_zstd_thin", "unrar"]
 use_zlib = ["flate2/zlib", "gzp/deflate_zlib", "zip/deflate-zlib"]
 use_zstd_thin = ["zstd/thin"]
+allow_piped_choice = []
 
 # For generating binaries for releases
 [profile.release]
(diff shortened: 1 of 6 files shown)


vrmiguel:
Thank you @talis-fb !! Great work! ❤️

PR #782, opened 2025-03-31
chore: Disable smart unpack when `--dir` is provided

### LINKED ISSUE #322: Disable smart unpack when `--dir` is provided
When decompressing an archive, there is a risk of messing up the current folder, if the archive is huge it will mix your folder with a lot of files and you'll have a hard time trying to clean it up afterwards.

To solve this, Ouch has a hidden feature called "smart unpack", when it detects that an archive has more than one file, it creates a new folder and unpacks all the contents inside of that folder, so cleaning up is as simple as deleting that folder.

However, if the user specified a target directory with `--dir`, we could (?) assume they intentionally want to decompress it into that folder directly, so "smart unpack" can be disabled, it will just create an unwanted nested folder.

talis-fb:
closes #322 

## Before, with Smart Unpack when `--dir` is provided:
image

## Now:
image

## When there are some conflict:
image

**NOTE**: When smart unpack is disabled the `.tmp-ouch-xxx` folder is no longer used inside `--dir` folder to hold unpacked files before renaming them. Instead, files are unpacked directly into the `--dir` folder.

### DIFF
--- a/src/commands/decompress.rs
+++ b/src/commands/decompress.rs
@@ -313,7 +313,7 @@ fn unpack(
     output_dir: &Path,
     question_policy: QuestionPolicy,
 ) -> crate::Result<ControlFlow<(), usize>> {
-    let is_valid_output_dir = !output_dir.exists() || (output_dir.is_dir() && output_dir.read_dir()?.count() > 0);
+    let is_valid_output_dir = !output_dir.exists() || (output_dir.is_dir() && output_dir.read_dir()?.count() == 0);
 
     let output_dir_cleaned = if is_valid_output_dir {
         output_dir.to_owned()
--- a/src/commands/decompress.rs
+++ b/src/commands/decompress.rs
@@ -298,31 +298,30 @@ fn execute_decompression(
     is_output_dir_explicit: bool,
 ) -> crate::Result<ControlFlow<(), usize>> {
     if is_output_dir_explicit {
-        return unpack(unpack_fn, output_dir, question_policy);
+        unpack(unpack_fn, output_dir, question_policy)
     } else {
-        return smart_unpack(
-            unpack_fn,
-            output_dir,
-            output_file_path,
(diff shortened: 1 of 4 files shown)


src/commands/decompress.rs
```diff
@@ -284,6 +290,49 @@ pub fn decompress_file(options: DecompressOptions) -> crate::Result<()> {
...
+/// Unpacks an archive creating the output directory, this function will create the output_dir
+/// directory or replace it if it already exists. The `output_dir` needs to be empty
+/// - If `output_dir` does not exist OR is a empty directory, it will unpack there
+/// - If `output_dir` exist OR is a directory not empty, the user will be asked what to do
+fn unpack(
+    unpack_fn: impl FnOnce(&Path) -> crate::Result<usize>,
+    output_dir: &Path,
+    question_policy: QuestionPolicy,
+) -> crate::Result<ControlFlow<(), usize>> {
+    let is_valid_output_dir = !output_dir.exists() || (output_dir.is_dir() && output_dir.read_dir()?.count() == 0);
```
marcospb19, line 316: `.read_dir()?.count()` reads all elements in a directory.

Can you check `.count()` calls in the project and optimize them.

Should be `.next().is_none()`, or, when it's not compared to `0`, `.nth(N).is_some/none()`

src/commands/decompress.rs
```diff
@@ -31,6 +31,7 @@ pub struct DecompressOptions<'a> {
     pub formats: Vec<Extension>,
     pub output_dir: &'a Path,
     pub output_file_path: PathBuf,
+    pub is_output_dir_explicit: bool,
```
marcospb19, line 34: Can you call this `is_output_dir_provided` or `is_dir_flag_present` or `has_dir_flag`

marcospb19 (changes requested):

PR #798, opened 2025-04-18
Merge folders in decompression

### LINKED ISSUE #466: Overwrite behavior removes files differently than expected
### Version

0.4.1

### Description

```
```

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Additional Information

_No response_

### LINKED ISSUE #727: Decompressing multiple archives conflicts with overwrite prompt
### Version

0.5.1

### Description

When decompressing multiple archives that contain the same file or directory names, ouch asks the user if the files should be overwritten multiple times, which can also be a random archive in the list:

### Current Behavior

> 
> ouch decompress testA.zip testB.zip testC.zip
> [INFO] Created temporary directory /tmp/experiment/./.tmpBER78Y to hold decompressed elements.
> [INFO] File 0 extracted to "/tmp/experiment/./.tmpBER78Y/test/"
> [INFO] Created temporary directory /tmp/experiment/./.tmpt6Uh2b to hold decompressed elements.
> [INFO] ".tmpBER78Y/test/a" extracted. (4.00 B)
> [INFO] Created temporary directory /tmp/experiment/./.tmp8oJZxo to hold decompressed elements.
> [INFO] File 0 extracted to "/tmp/experiment/./.tmpt6Uh2b/test/"
> [INFO] Successfully moved /tmp/experiment/./.tmpBER78Y/test to ./test.
> [INFO] File 0 extracted to "/tmp/experiment/./.tmp8oJZxo/test/"
> [INFO] Successfully decompressed archive in current directory (2 files).
> [INFO] ".tmpt6Uh2b/test/b" extracted. (4.00 B)
> [INFO] ".tmp8oJZxo/test/c" extracted. (4.00 B)
> Do you want to overwrite './test'? [Y/n] Do you want to overwrite './test'? [Y/n]

### Expected Behavior

The user should be prompted multiple times, one after another with specification of which archive is asking to overwrite the specified files.

Example:

> Archive testb.zip: Do you want to overwrite './test'? [Y/n] n
> Archive testC.zip: Do you want to overwrite './test'? [Y/n] Y

### Additional Information

_No response_

tommady:
try to
close #466
close #727

I am very open-minded to accepting any suggestions.
The image below is just an example of fulfilling the purpose of those issues.

image


src/utils/question.rs
```diff
@@ -48,49 +48,71 @@ pub enum FileConflitOperation {
     /// Rename the file
     /// It'll be put "_1" at the end of the filename or "_2","_3","_4".. if already exists
     Rename,
+    /// Merge duplicated files
```
marcospb19, line 51:
```suggestion
    /// Merge conflicting folders.
```

src/utils/question.rs
```diff
@@ -48,49 +48,71 @@ pub enum FileConflitOperation {
...
+    Merge,
```
marcospb19, line 52: We have a flag for always overwriting and a flag for always renaming, can you add a flag `--merge` to always merge? (without asking the prompt).

marcospb19 (changes requested):
Thanks for this PR! I never before considered this could be an option in the prompt instead of the default, it seems like a good idea so I'm thinking about letting that be.

Can you add a test that unpacks two archives with the following structures:
```
a/b/c/1
```
and
```
a/b/c/2
```

And show that when decompressed and provided `m` via stdin, they will merge into:

```
a/
 b/
  c/
   1
   2
```

You can start from a copy of `multiple_files_with_conflict_and_choice_to_rename` that uses the 'allow_piped_choice' feature flag (which we should someday turn into a flag), with that you can pass the stdin choice you want.

marcospb19 (changes requested):
This is also on a good path for merging :) so don't forget a changelog entry, I think the PR title reflects this change well.

tommady:
> Thanks for this PR! I never before considered this could be an option in the prompt instead of the default, it seems like a good idea so I'm thinking about letting that be.
> 
> Can you add a test that unpacks two archives with the following structures:
> 
> ```
> a/b/c/1
> ```
> 
> and
> 
> ```
> a/b/c/2
> ```
> 
> And show that when decompressed and provided `m` via stdin, they will merge into:
> 
> ```
> a/
>  b/
>   c/
>    1
>    2
> ```
> 
> You can start from a copy of `multiple_files_with_conflict_and_choice_to_rename` that uses the 'allow_piped_choice' feature flag (which we should someday turn into a flag), with that you can pass the stdin choice you want.

I am stuck on how to write the test. Thanks for your guidance.

src/utils/question.rs
tommady: Could you guide me on which flag represents the always renaming?
because, from my understanding
```bash
flag --yes = always override
```
And yes, I can add a new flag in the decompress command to always merge.

marcospb19: Huh I thought the rename PR added a flag for that...

tests/integration.rs
```diff
@@ -604,3 +604,67 @@ fn no_git_folder_after_decompression_with_gitignore_flag_active() {
...
+
+    fs::remove_dir_all(&source_path)?;
+    fs::create_dir_all(&source_path)?;
+    let archive1 = root_path.join(format!("archive1.{}", merge_extensions(&ext, &extra_extensions)));
+    crate::utils::cargo_bin()
+        .arg("compress")
+        .args([
+            fs::File::create(source_path.join("file4.txt"))?.path(),
+            fs::File::create(source_path.join("file5.txt"))?.path(),
+            fs::File::create(source_path.join("file6.txt"))?.path(),
```
marcospb19, line 643: What happens if you change the two archives to be:

`1 2 3`
and 
`3 4 5` (don't have the 6th)

Does it ask to overwrite again?

tommady: I updated the test case, and it does not ask for overwriting again.

src/utils/question.rs
tommady: hmmm, sorry, I read the question.rs deeply found out to add a new flag --merge we need to refactor the function `pub fn user_wants_to_overwrite` which might be beyond the PR's scope.

tests/integration.rs
marcospb19: I don't think this is ideal, after telling Ouch to merge folders, I would expect it to prompt me again about conflicts, or, we can have folder merging enabled by default and then receive the usual prompts for conflicts asking if I want to overwrite or not.

For reference, the latter is how the 7z GUI works.

tommady: hmmm, yes, it is possible, but it needs to refactor the question.rs file very hard.
you can put me in another PR to do that.
WDYT.

marcospb19: Cool :) 

I'd love to see another PR for that but don't feel too pressured, I'd understand if you're not able to put the time

marcospb19 (approved):
looking good

We decided that https://github.com/ouch-org/ouch/pull/798#discussion_r2063809312 can be addressed later.

src/utils/question.rs
marcospb19: I think these would be a good addition, these would be welcome for another PR :) primarily the `--merge`, right now we have `--overwrite` in the form of `--yes`, which I think should be adjusted now that we have more than `yes` and `no` questions.

I think `--yes` shouldn't overwrite, so for folks to use this in scripts, they can provide `--yes` for the `y/n` questions and `--overwrite` for file conflicts.

marcospb19:
Hey, I found a bug around this implementation:

```
$ mkdir parent/
$ touch parent/a
$ ouch c parent a.tar
[INFO] Compressing 'parent'
[INFO] Compressing 'parent/a'
[INFO] Successfully compressed 'a.tar'
$ mv parent/a parent/b
$ ouch c parent b.zip
[INFO] Compressing 'parent'
[INFO] Compressing 'parent/b'
[INFO] Successfully compressed 'b.zip'
$ rm parent -fr
$
$ ouch d a.tar
[INFO] Created temporary directory /home/marcospb19/./tmp-ouch-dhhuxX to hold decompressed elements
[INFO] extracted (  0.00   B) "tmp-ouch-dhhuxX/parent"
[INFO] extracted (  0.00   B) "tmp-ouch-dhhuxX/parent/a"
[INFO] Successfully moved "/home/marcospb19/./tmp-ouch-dhhuxX/parent" to "./parent"
[INFO] Successfully decompressed archive in current directory
[INFO] Files unpacked: 2
$ ouch d b.zip
[INFO] Created temporary directory /home/marcospb19/./tmp-ouch-ECtvGC to hold decompressed elements
[INFO] File 0 extracted to "/home/marcospb19/./tmp-ouch-ECtvGC/parent/"
[INFO] extracted (  0.00   B) "tmp-ouch-ECtvGC/parent/b"
Do you want to overwrite ./parent? [y/n/r/m]
m
[ERROR] failed to rename file from /home/marcospb19/./tmp-ouch-ECtvGC/parent to ./parent
```

Looks like a scenario we forgot to test.

tommady:
thank you, I created an [issue](https://github.com/ouch-org/ouch/issues/825) and I'll fix it.

PR #800, opened 2025-04-18
Rename short flag `--threads` from `-c` to `-t`

marcospb19:


marcospb19:
note for future self: following what compilers do we can add `-j` as an alias

marcospb19:
looks like `zstd_encoder.multithread` isn't respecting this setting

Issue #813, opened 2025-04-29
Almost deleted my home directory


salahoued:
It my seem weird using `-d ./` because `ouch` extract in the current directory already, I'm not sure what I was trying to do back then.. :(

But when I did run `ouch d -y -d ./ Downloads/Compressed/<file>`  almost deleted my HOME directory, and I'm not sure why some other directories and files weren't deleted.
What I observed is `.config/*` `.cargo` `.var` `Musique` some file in `$HOME` were deleted and `.local` `.cache` `Document` `Images` a `so-config` where I `stowed` my dotfiles and some other files weren't.

marcospb19:
So sorry to hear that, when running with `-d .` it definitely shouldn't try replacing the current dir, I don't see why it would try that.

Thanks for reporting!

N4taaa:
Just lost a bunch of files in a directory the same way, ouch! 😅 The things in life that teach you should back up (too late).
I'm curious, does the replacing the directory effect of the `--dir` flag have any use case that warrants it being the default behavior?

marcospb19:
Some changes are being made in the main branch, we're adding an option to merge folders instead of overwriting them.

The code will ask on every conflict whether you want to overwrite or merge.

How can we improve that? Should it merge by default on folder conflicts, and only ask for overwriting on file conflicts?

Maybe we remove `-y` and `yes` being the "overwrite" option as people are inclined to just type 'y' without thinking much about it.

Instead of 'y/n' maybe we could ask `m` for merge, `o` for overwrite, `s` for skip and `q` for quit??

Maybe simpler, should Ouch always merge to the folder where `--dir` is provided, and then, ask for each conflict?

Thoughts on this? I need more input to come up with something decent. Ideally I don't have to change this more than once.

NiceGuyIT:
I was bit by this bug and it removed `~/.local/share`. 

> How can we improve that? Should it merge by default on folder conflicts, and only ask for overwriting on file conflicts?
>
> Should it also merge to the top-level folder where `--dir` is provided, and then, ask for each conflict?

From someone who has been working with software for decades, here's what makes sense to me.

1. Clearly indicate breaking changes. The releases page is used to indicate what changed. That's great! It doesn't clearly indicate breaking changes. Perhaps prefixing breaking changes with BREAKING: to indicate it's a breaking change.
[...]

PR #821, opened 2025-05-13
decompress: Fix --dir erasing the whole folder

### LINKED ISSUE #820: Regression: disabling smart unpack causes Ouch to delete the whole directory
### Version

0.6.1

### Description

https://github.com/ouch-org/ouch/pull/782 introduced unpacking straight into the target directory when `--dir` is specified.

In Windows, however, this does not work as expected. It tries deleting and recreating the target directory instead of just writing the files.

### Current Behavior

```
❯ ouch decompress -d "$env:TEMP" D:\Downloads\libclang-20.1.2.tar.xz
Do you want to overwrite C:\Users\Amalia\AppData\Local\Temp\? [y/n/r]
```

(obviously `n`)

### Expected Behavior

```
❯ ouch decompress -d "$env:TEMP" D:\Downloads\libclang-20.1.2.tar.xz
[INFO] "Users\\Amalia\\AppData\\Local\\Temp\\libclang-20.1.2" extracted. (0 B)
[INFO] "Users\\Amalia\\AppData\\Local\\Temp\\libclang-20.1.2/LICENSE.TXT" extracted. (15.14 kiB)
[INFO] "Users\\Amalia\\AppData\\Local\\Temp\\libclang-20.1.2/x64" extracted. (0 B)
[INFO] "Users\\Amalia\\AppData\\Local\\Temp\\libclang-20.1.2/x64/bin" extracted. (0 B)
[INFO] "Users\\v\\AppData\\Local\\Temp\\libclang-20.1.2/x64/bin/libclang.dll" extracted. (80.14 MiB)
[INFO] "Users\\Amalia\\AppData\\Local\\Temp\\libclang-20.1.2/x86" extracted. (0 B)
[INFO] "Users\\Amalia\\AppData\\Local\\Temp\\libclang-20.1.2/x86/bin" extracted. (0 B)
[INFO] "Users\\Amalia\\AppData\\Local\\Temp\\libclang-20.1.2/x86/bin/libclang.dll" extracted. (75.06 MiB)
[INFO] Successfully decompressed archive in C:\Users\Amalia\AppData\Local\Temp.
[INFO] Files unpacked: 8
```

### Additional Information

Compare output of https://gitlab.freedesktop.org/amyspark/gstreamer-rs/-/jobs/74156023 to https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/jobs/76236804#L5522.

amyspark:
Fixes #820

### DIFF
--- a/src/commands/decompress.rs
+++ b/src/commands/decompress.rs
@@ -338,7 +338,7 @@ fn unpack(
     output_dir: &Path,
     question_policy: QuestionPolicy,
 ) -> crate::Result<ControlFlow<(), usize>> {
-    let is_valid_output_dir = !output_dir.exists() || (output_dir.is_dir() && output_dir.read_dir()?.next().is_none());
+    let is_valid_output_dir = !output_dir.exists() || output_dir.is_dir();
 
     let output_dir_cleaned = if is_valid_output_dir {
         output_dir.to_owned()


amyspark:
From the test failures, it seems that disabling smart unpacking assumes one would be willing to clear the directory. This doesn't seem consistent with any other unpackers, e.g. `tar` where a change of folder just means setting where one wants the payloads written to.

marcospb19:
Hi! If I understand, this enables Ouch to unpack something into an existing non-empty directory, makes sense.

This is related to https://github.com/ouch-org/ouch/pull/798, which implements merging folders (that wasn't a thing before), I considered making Ouch `merge` by default, and then you get asked prompts about further conflicts.

But I still haven't figured out what is the most intuitive, should merging only happen by default if `--dir` is provided? (If so, this PR is also merging when `--dir` is not provided.)

Do you have any thoughts on that? Would it be surprising if `ouch decompress archive.tar` unpacked files into existing folders? That could be hard to reverse if it wasn't intended.

amyspark:
> Would it be surprising if ouch decompress archive.tar unpacked files into existing folders?

It was surprising for me that `ouch decompress X.tar` deletes a folder before unpacking on it. Tools like `tar`, 7-Zip, `unzip` etc. all decompress in the target directory with no issue, just printing out which file they're writing or overwriting.

Had I ran into this on my desktop instead of on CI, it could've destroyed my whole user directory, hence why I'd like to see this fixed.

marcospb19:
Hey, I'm going to close this for now as I think I first need to improve the default behavior to merge folders by default instead of deleting them, then, I'll make sure the scenario you mentioned is covered.

Related to, and tracked at https://github.com/ouch-org/ouch/issues/813.

Thanks tho!!

PR #826, opened 2025-05-18
Fix Unpacking with merge flag failed without --dir flag

### LINKED ISSUE #825: Unpacking with merge flag failed without --dir flag
### Version

v0.6.1

### Description

https://github.com/ouch-org/ouch/pull/798#issuecomment-2888477955

### Current Behavior

```bash
[ERROR] failed to rename file from /home/arch/code/xxx/ouch/./tmp-ouch-cIZ8w1/parent to ./parent
```

### Expected Behavior

No error occurs

### Additional Information

by this [PR](https://github.com/ouch-org/ouch/pull/798) 
We introduced a merge action for the decompress command, but were missing some test cases that would lead to a failure.

tommady:
close #825

### DIFF
--- a/src/utils/fs.rs
+++ b/src/utils/fs.rs
@@ -209,3 +209,16 @@ pub fn try_infer_extension(path: &Path) -> Option<Extension> {
         None
     }
 }
+
+pub fn copy_dir(src: &Path, dst: &Path) -> crate::Result<()> {
+    for entry in fs::read_dir(src)? {
+        let entry = entry?;
+        let ty = entry.file_type()?;
+        if ty.is_dir() {
+            copy_dir(entry.path().as_path(), dst.join(entry.file_name()).as_path())?;
+        } else {
+            fs::copy(entry.path(), dst.join(entry.file_name()).as_path())?;
+        }
+    }
+    Ok(())
+}
--- a/src/utils/fs.rs
+++ b/src/utils/fs.rs
@@ -212,6 +212,12 @@ pub fn try_infer_extension(path: &Path) -> Option<Extension> {
 
 /// Copy the src directory into the dst directory recursively
 pub fn copy_dir(src: &Path, dst: &Path) -> crate::Result<()> {
+    if !src.exists() || !dst.exists() {
+        return Err(crate::Error::NotFound {
+            error_title: "source or destination directory does not exist".to_string(),
+        });
+    }
+
     for entry in fs::read_dir(src)? {
         let entry = entry?;
         let ty = entry.file_type()?;
--- a/src/utils/fs.rs
+++ b/src/utils/fs.rs
@@ -210,6 +210,7 @@ pub fn try_infer_extension(path: &Path) -> Option<Extension> {
     }
 }
 
+/// Copy the src directory into the dst directory recursively
 pub fn copy_dir(src: &Path, dst: &Path) -> crate::Result<()> {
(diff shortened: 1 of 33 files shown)


marcospb19:
Hey, apologies for the huge delay, I took a break. This is marked as a draft, is it ready for review now? :)

tommady:
ok, right now ignore the Clippy error, the update test case passed...
```bash
test unpack_multiple_sources_into_the_same_destination_with_merge ... ok
```

marcospb19:
Hah, did you expect it to fail but it's passing?

I triggered a re-run just in case :eyes:

marcospb19:
> But the integration test cannot simulate this issue

@tommady I figured the tests don't cover this bug because the bug itself only happens when the `--dir` or `-d` flag is not present, and all the tests rely on this `-d` flag so that it doesn't pollute the current dir with test artifacts, but instead, write them to `/tmp`.

Unfortunately, due to the way cargo test runner works, we also can't use `std::env::set_current_dir` to avoid this.

So there are 2 ways forward I can think of:

1. Not having tests that cover this specific bug when `-d` is not present.
2. Accepting test artifacts to current folder.

I'm fine with 1 I guess, what do you think?

tommady:
> > But the integration test cannot simulate this issue
> 
> @tommady I figured the tests don't cover this bug because the bug itself only happens when the `--dir` or `-d` flag is not present, and all the tests rely on this `-d` flag so that it doesn't pollute the current dir with test artifacts, but instead, write them to `/tmp`.
> 
> Unfortunately, due to the way cargo test runner works, we also can't use `std::env::set_current_dir` to avoid this.
> 
> So there are 2 ways forward I can think of:
> 
> 1. Not having tests that cover this specific bug when `-d` is not present.
> 2. Accepting test artifacts to current folder.
> 
> I'm fine with 1 I guess, what do you think?

interesting.
anyway thanks for your explanation
now I see this issue more clearly, thanks!

I'll try to do the option 2 *Accepting test artifacts to current folder*
thanks.

tommady:
hi @marcospb19 
could you kindly do a review while you have time?
thanks.

src/utils/fs.rs
```diff
@@ -209,3 +209,23 @@ pub fn try_infer_extension(path: &Path) -> Option<Extension> {
...
+        });
+    }
+
+    for entry in fs::read_dir(src)? {
+        let entry = entry?;
+        let ty = entry.file_type()?;
+        if ty.is_dir() {
+            copy_dir(entry.path().as_path(), dst.join(entry.file_name()).as_path())?;
+        } else {
+            fs::copy(entry.path(), dst.join(entry.file_name()).as_path())?;
```
marcospb19, line 227: The only problem I see here is, this is duplicating everything that's already written to the disk, so an archive of 20GB can't be decompressed in a filesystem with only 15GB of free space (unless FS copy-on-write is enabled).

Instead of `cp` we should do a `mv`.

Can you change from `fs::copy` to `fs::rename` to achieve that?

(There is a caveat that this doesn't work well for crossing filesystem boundaries but that's a problem for future me.)

marcospb19 (changes requested):

marcospb19 (approved):
Great! Thanks for sticking to this, really appreciate it.

marcospb19:
Hey @tommady if you have an email I can msg you, could you ping me at marcospb19 @ hotmail.com

Thanks :)

PR #834, opened 2025-06-28
Support for comic book aliases

md9753:
Addresses #831

[Several archive formats have commonly-used aliases for storing comic books.](https://en.wikipedia.org/wiki/Comic_book_archive) This PR adds those aliases for the formats Ouch already supports:
- tar: .cbt
- zip: .cbz
- 7z: .cb7
- rar: .cbr

Screenshot_20250628_112935

I'm very new to Rust, so a close eye when looking this over would be appreciated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Categories Used:
 
 ### New Features
 
+- Add aliases for comic books :)
 - Merge folders in decompression [\#798](https://github.com/ouch-org/ouch/pull/798) ([tommady](https://github.com/tommady))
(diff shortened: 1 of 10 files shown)


md9753:
I'm still learning about git -- looks like this is keeping the individual commit names on my branch? I was under the impression github squashed everything. I'll reopen this after renaming the commits to be more informative

md9753:
I was reading some tutorials and came across that last "not ideal" method, so I squashed everything locally and force-pushed to the fork I made. Then I couldn't edit this PR to reflect those changes, so after reading more and still not being able to fix it, I opened #835.

I don't know if I should try to fix this PR since #835 exists. Rather than making another assumption, let me know how to proceed and I'll do my best to make it happen.

marcospb19:
> I squashed everything locally and force-pushed to the fork I made. Then I couldn't edit this PR to reflect those changes

That should've worked!! weird, maybe you force-pushed to the wrong branch or something like that

Issue #839, opened 2025-07-05
Error while decompressing .zip with ouch, but default OS GUI archiver in Linux Mint decompress OK


yh-sb:
### Version

0.6.1

### Description

Download this .zip archive: https://github.com/balena-io/etcher/releases/download/v2.1.2/balenaEtcher-linux-x64-2.1.2.zip

Try to decompress with ouch:
```sh
ouch d balenaEtcher-linux-x64-2.1.2.zip
```
See error: `[ERROR] No such file or directory (os error 2) - File not found`

Try to unpack this .zip with default OS file manager (Nemo on Linux mint 21.3).
See **successful unpack without any errors**.

### Current Behavior

Error while unpacking .zip:
```log
[INFO] Created temporary directory /home/y/Downloads/./tmp-ouch-MtcctU to hold decompressed elements
[INFO] File 0 extracted to "/home/y/Downloads/./tmp-ouch-MtcctU/balenaEtcher-linux-x64/"
[INFO] extracted ( 10.34 MiB) "tmp-ouch-MtcctU/balenaEtcher-linux-x64/LICENSES.chromium.html"
[INFO] extracted (  7.45 MiB) "tmp-ouch-MtcctU/balenaEtcher-linux-x64/libvulkan.so.1"
[INFO] extracted (  2.72 MiB) "tmp-ouch-MtcctU/balenaEtcher-linux-x64/libffmpeg.so"
[INFO] extracted (107.00   B) "tmp-ouch-MtcctU/balenaEtcher-linux-x64/vk_swiftshader_icd.json"
[INFO] extracted (  1.10 kiB) "tmp-ouch-MtcctU/balenaEtcher-linux-x64/LICENSE"
[INFO] extracted ( 54.25 kiB) "tmp-ouch-MtcctU/balenaEtcher-linux-x64/chrome-sandbox"
[INFO] extracted (  1.28 MiB) "tmp-ouch-MtcctU/balenaEtcher-linux-x64/chrome_crashpad_handler"
[INFO] File 8 extracted to "/home/y/Downloads/./tmp-ouch-MtcctU/balenaEtcher-linux-x64/resources/"
[INFO] extracted (  5.59 MiB) "tmp-ouch-MtcctU/balenaEtcher-linux-x64/resources/app.asar"
[INFO] extracted (490.00   B) "tmp-ouch-MtcctU/balenaEtcher-linux-x64/resources/sudo-askpass.osascript-en.js"
[...]

Issue #822, opened 2025-05-15
Backslashes when creating a zip


Tyarel8:
### Version

0.6.1

### Description

When creating a `.zip` file in windows, the path of folders use backslashes, which results in errors when opening the archive with apps like 7zip. It shows all the path as the filename with weird characters for path separators.

![Image](https://github.com/user-attachments/assets/cf50e763-e42e-4f30-804e-d827555307ec)

Other formats like `7z` work normal.

```
→ ouch list test.7z
Archive: \\?\C:\Users\<user>\Downloads\test.7z
a/b.txt
a/

→ ouch list test.zip
Archive: \\?\C:\Users\<user>\Downloads\test.zip
a//
a\b.txt
```

marcospb19:
Thanks for reporting, I do not own a Windows machine so support is subpar.

Issue #807, opened 2025-04-20
`ouch list` output duplicates slashes when STDOUT is piped


marcospb19:
### Version

0.6.0

### Description

When I run

```
$ ouch l ouch-aarch64-pc-windows-msvc.zip | head -n 5
Archive: /home/marcospb19/Downloads/ouch-releases/ouch-aarch64-pc-windows-msvc.zip
ouch-aarch64-pc-windows-msvc//
ouch-aarch64-pc-windows-msvc/man//
ouch-aarch64-pc-windows-msvc/man/ouch-decompress.1
ouch-aarch64-pc-windows-msvc/man/ouch-compress.1
```

I see duplicated slashes when it's a directory.

marcospb19:
This archive was created with `zip -r output input`.

When I recreate it with `Ouch` itself, it doesn't do that.

Could be that CLI `zip` adds a `/` at the path, but we are also adding a `/` when outputting.

tommady:
This issue seems unable to be reproduced on linux os.
And I am sorry I don't have windows os lol.

![Image](https://github.com/user-attachments/assets/f767cc73-327e-425b-8281-9ada24a6ebd7)

marcospb19:
Hey @tommady that'd only happen if the output is piped so you'd need to change `ouch l a.zip` to `ouch l a.zip | head -n 5` :) .

mochienya:
This issue still exists for me. Tested on a GitHub tarball with both fish and bash, ouch 0.6.1 from nixpkgs, x86_64-linux on NixOS. The issue was never Windows specific from what I can tell either: the zip file @marcospb19 used to test it happened to be an ouch archive for Windows, but if you look at the `Archive:` line, it's a UNIX path.

<img width="785" height="337" alt="output of `ouch ls` in a terminal" src="https://github.com/user-attachments/assets/7c43aced-a072-4331-9c4d-514851a65ac4" />

<img width="677" height="337" alt="output of `ouch ls` piped into `cat` in a terminal. the directories all have 2 trailing slashes" src="https://github.com/user-attachments/assets/19f9333f-803a-43bd-9e33-66217968d618" />

marcospb19:
I opened #952 to fix this, but it was already fixed by another commit, likely it was #934 from @Tommady.

Closing as solved!

PR #841, opened 2025-07-07
update zip to 4.2.0 and fix some bugs

zzzsyyy:
Should close:
* #839 
* #717
* #842 
* #822
* #807 

+ updated the zip crate from v0.6.6 to v4.2.0

+ added detailed logging for symlink creation (`linking source -> target`), don't set_last_modified_time and unix_set_permissions when file is a symlink

+ fix duplicate trailing slashes in no-color mode

+ fix backslashes issue when creating zip on Windows (tested)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,12 +36,14 @@ Categories Used:
 - Give better error messages when archive extensions are invalid [\#817](https://github.com/ouch-org/ouch/pull/817) ([marcospb19](https://github.com/marcospb19))
 - Add aliases for `--password` flag (`--pass` and `--pw`) [\#847](https://github.com/ouch-org/ouch/pull/847) ([marcospb19](https://github.com/marcospb19))
 - Use `lzma-rust2` crate instead of `liblzma` crate [\#867](https://github.com/ouch-org/ouch/pull/867) ([sorairolake](https://github.com/sorairolake))
+- Avoid loading entire 7z archive into memory when listing [\#860](https://github.com/ouch-org/ouch/pull/860) ([86xsk](https://github.com/86xsk))
 
 ### Bug Fixes
 
 - Fix tar extraction count when --quiet [\#824](https://github.com/ouch-org/ouch/pull/824) ([marcospb19](https://github.com/marcospb19))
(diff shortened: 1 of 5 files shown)


src/error.rs
```diff
@@ -216,11 +216,17 @@ impl From<zip::result::ZipError> for Error {
         use zip::result::ZipError;
         match err {
             ZipError::Io(io_err) => Self::from(io_err),
-            ZipError::InvalidArchive(filename) => Self::InvalidZipArchive(filename),
+            ZipError::InvalidArchive(filename) => Self::InvalidZipArchive(filename.to_string()),
             ZipError::FileNotFound => Self::Custom {
                 reason: FinalError::with_title("Unexpected error in zip archive").detail("File not found"),
             },
-            ZipError::UnsupportedArchive(filename) => Self::UnsupportedZipArchive(filename),
+            ZipError::UnsupportedArchive(filename) => Self::UnsupportedZipArchive(filename.to_string()),
```
vrmiguel: It does seem like we could still use &'static str here

vrmiguel:
There was a bit of overlap with #850:


```
        match file.name().ends_with('/') {
            _is_dir @ true => {
                // This is printed for every file in the archive and has little
                // importance for most users, but would generate lots of
                // spoken text for users using screen readers, braille displays
                // and so on
                if !quiet {
                    info(format!("File {} extracted to \"{}\"", idx, file_path.display()));
                }
<<<<<<< fix/zip
                fs::create_dir_all(&file_path)?;

                #[cfg(unix)]
                unix_set_permissions(&file_path, &file)?;
=======

                let mode = file.unix_mode();
                let is_symlink = mode.is_some_and(|mode| mode & 0o170000 == 0o120000);

                if is_symlink {
                    let mut target = String::new();
                    file.read_to_string(&mut target)?;

                    #[cfg(unix)]
                    std::os::unix::fs::symlink(&target, &file_path)?;
                    #[cfg(windows)]
                    std::os::windows::fs::symlink_dir(&target, file_path)?;
                } else {
                    fs::create_dir_all(&file_path)?;
                }
>>>>>>> main
```

Issue #844, opened 2025-07-10
softlink is not preserved after tar.gz


demin-han:
### Version

0.6.1

### Description

directory tree:
test-softlink
|-- aa
     |-- aa.log
|-- bb -> aa    // bb is a softlink to aa

compress:
ouch compress test-softlink test-softlink.tar.gz

after decompress:
ouch decompress test-softlink.tar.gz

**bb will not link to aa.**

demin-han:
ouch compress test-link ouch-test-link.tar.gz
[ouch-test-link.tar.gz](https://github.com/user-attachments/files/21178214/ouch-test-link.tar.gz)
tar czf tar-test-link.tar.gz test-link
[tar-test-link.tar.gz](https://github.com/user-attachments/files/21178213/tar-test-link.tar.gz)

@zzzsyyy please check.

tommady:
hmmm I cannot reproduce either.

<img width="1110" height="1590" alt="Image" src="https://github.com/user-attachments/assets/ede4cdcc-b165-4646-8bc7-2d9763366342" />

tommady:
hmmm, weird, still can't reproduce 

<img width="1106" height="1363" alt="Image" src="https://github.com/user-attachments/assets/4ef0b467-5203-4606-9c18-0e80969ab922" />

demin-han:
> hmmm, weird, still can't reproduce
> 

the bb link in your picture is a bad link and shown in red color.

you should execute ls -sf aa bb in test-softlink dir.

tommady:
Yes, I can reproduce it now and will try to fix it. thanks

PR #850, opened 2025-07-17
Fix folder softlink is not preserved after packing

tommady:
close: #844

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ Categories Used:
 - Give better error messages when archive extensions are invalid [\#817](https://github.com/ouch-org/ouch/pull/817) ([marcospb19](https://github.com/marcospb19))
 - Add aliases for `--password` flag (`--pass` and `--pw`) [\#847](https://github.com/ouch-org/ouch/pull/847) ([marcospb19](https://github.com/marcospb19))
 - Use `lzma-rust2` crate instead of `liblzma` crate [\#867](https://github.com/ouch-org/ouch/pull/867) ([sorairolake](https://github.com/sorairolake))
+- Avoid loading entire 7z archive into memory when listing [\#860](https://github.com/ouch-org/ouch/pull/860) ([86xsk](https://github.com/86xsk))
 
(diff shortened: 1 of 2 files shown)


tommady:
cc @demin-han

image

tommady:
hi @marcospb19 
could you kindly do a review while you have time?
thanks.

src/archive/tar.rs
```diff
@@ -40,6 +40,10 @@ pub fn unpack_archive(reader: Box<dyn Read>, output_folder: &Path, quiet: bool)
 
                 #[cfg(unix)]
                 std::os::unix::fs::symlink(&target, &full_path)?;
+
+                // FIXME: how to detect whether the destination is a folder or a regular file?
+                // regular file should use fs::symlink_file
+                // folder should use fs::symlink_file
```
marcospb19, line 46: you meant
```suggestion
                // folder should use fs::symlink_dir
```

src/archive/tar.rs
```diff
@@ -141,9 +145,7 @@ where
                 info(format!("Compressing '{}'", EscapedPathDisplay::new(path)));
             }
 
-            if path.is_dir() {
-                builder.append_dir(path, path)?;
-            } else if path.is_symlink() && !follow_symlinks {
+            if ((path.is_dir() && path.symlink_metadata()?.is_symlink()) || path.is_symlink()) && !follow_symlinks {
```
marcospb19, line 148: The `path.is_dir() && path.symlink_metadata()?.is_symlink()` part seem to be redundant considering it's followed by `path.is_symlink()`.

You could cover both cases with 'is_symlink'

However, that function fails silently so, my suggestion is sticking to symlink_metadata to catch errors:

```suggestion
            if !follow_symlinks && path.symlink_metadata()?.is_symlink() {
```

I think that covers everything you wanted, and short-circuits before doing IO when follow_symlinks is true, makes sense?

tommady: correct! I'd like your review style, which focuses on the performance.

Issue #843, opened 2025-07-09
Adjust `[INFO]` color


lomirus:
Currently, the `[INFO]` output is in yellow.

<img width="1294" height="270" alt="Image" src="https://github.com/user-attachments/assets/e498988d-9415-441f-9739-ce3aff7f933e" />

However, in UI design, yellow is often used to indicate warnings. The first time I used ouch, seeing the yellow `[INFO]` made me mistakenly think the archive was corrupted. It wasn't until I read carefully that I realized there was no issue.

Referring to other tools in the Rust ecosystem (such as `log` and `cargo`), I suggest changing `[INFO]` to green, with white as the second-best option.

PR #858, opened 2025-08-21
fix:  adjust INFO color to green

etolbakov:
close: https://github.com/ouch-org/ouch/issues/843


etolbakov:
@marcospb19 
Could you please take a look when you have a minute?

marcospb19 (approved):
Ty! :)

PR #907, opened 2026-02-15
Breaking change: remove auto-flattening feature

marcospb19:
Previously called "smart unpack", all this feature did was automatically flatten the top-level directory of an archive when decompressing an archive with a single element in its root

It's being removed right now for being unpredictable, might be re-added in the future with flags or config file, basically, something that makes it opt-in and not the default


FalseDev:
I quite liked this feature, any flag or env variable to enable it would be appreciated

marcospb19:
@FalseDev I see, we're now discussing a better approach in #963.

So, nothing right now, hopefully soon.

PR #931, opened 2026-03-02
feat: add --completions flag for shell completion generation (1)

marcospb19:
Reopening PR #896.

(Branch for this PR was erased, but it had our license.)


figsoda:
Don't we already generate shell completions at build time? Are there any benefits to also do this at run time?

dimpen:
This would be really helpful. I came across this issue looking for completions for zsh.
I'm installing with cargo so I don't have completions.

For example in my setup I'm using zinit for zsh and it needs a way to get the completions `_ouch`.
Since there's no static completions file in the repo (like `completions/zsh`) the next best thing is to have `ouch --completions zsh`.

marcospb19:
@dimpen thanks, I'm reconsidering, provided that I manage to simplify this code and I don't see a huge binary size increase, I might include this in and switch away from build-time completions generation.

PR #934, opened 2026-03-05
list: show symlink targets (for tar and zip)

### LINKED ISSUE #932: `ouch list` should show symlink targets for supported formats
Take this `example.tar` example:

```sh
$ ouch l example.tar
Archive: "example.tar"
example
example/file
example/link

$ ouch l example.tar --tree
Archive: "example.tar"
└── example
   ├── file
   └── link
```

The expected output:

```sh
$ ouch l example.tar
Archive: "example.tar"
example
example/file
example/link -> target_here

$ ouch l example.tar --tree
Archive: "example.tar"
└── example
   ├── file
   └── link -> target_here
```

tommady:
close: #932


marcospb19 (changes requested):
Hey sorry for the eager review but what if is_dir and symlink_target get collapsed into one field for an enum for file type.

Regular and Directory unit variants
Symlink tuple variant with PathBuf inside.

so you express in the type system that `symlink_target` can't be some while is_dir is true.

wdyt

marcospb19 (changes requested):

src/list.rs
```diff
@@ -41,40 +49,58 @@ pub fn list_files(
...
+        }
+        FileType::Symlink { target } | FileType::Hardlink { target } => {
+            if is_running_in_accessible_mode() {
+                // Accessible mode: use "->" for screen readers
+                let _ = writeln!(out, "{} -> {}", name, target.display());
+            } else {
+                // Normal mode: use "->" with colors
+                let _ = writeln!(
+                    out,
+                    "{}{}{} -> {}{}{}",
```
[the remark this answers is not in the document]
marcospb19: Still prefer the `->` symbol for symlinks and `->` + `(hardlink)` :v sorry

Similar to what `tree -a` does in the CLI (the one I always download from pacman).

marcospb19:
Wait I accidentally pushed something directly to main :facepalm:, let me solve this conflict.

marcospb19 (approved):
lgtm!

src/archive/zip.rs
```diff
@@ -119,15 +119,31 @@ where
...
+        } else if let Some(target) = file
+            .unix_mode()
+            .filter(|mode| mode & 0o170000 == 0o120000)
+            .and_then(|_| {
+                let mut s = Vec::new();
+                file.read_to_end(&mut s)
+                    .ok()
+                    .map(|_| PathBuf::from(String::from_utf8_lossy(&s).into_owned()))
+            })
+        {
```
marcospb19, line 140: Oops, I just caught this piece.

Can you replace these by our `read_file_type` followed by Path::read_link?

marcospb19 (changes requested):

src/archive/zip.rs
marcospb19: Heads up I also pushed a small rename you have to pull 1 commit.

marcospb19: Oh also needs a changelog entry :D this is a good addition

tommady: > Oops, I just caught this piece.
> 
> Can you replace these by our `read_file_type` followed by Path::read_link?

I think we cannot do that because at this point we are reading virtual entries directly from the ZIP archive stream-the files havent't been extracted to disk yet.

marcospb19: oops, totally misunderstood when re-reading

so it's all good

marcospb19 (approved):
Yay thanks

For the changelog I'll later figure out an automation later so you don't have to do this manually xD.

Merging

PR #937, opened 2026-03-12
fix:unify directory extraction messages for zip and tar

venoosoo:
Both zip and tar now print the same message when extracting a directory:
`[INFO] Directory "..." created`

Also updated the insta snapshot to reflect the new output.

PR #944, opened 2026-03-20
fix: show "Directory ... created" instead of misleading byte size

### LINKED ISSUE #328: Improve unpacking output when creating an empty directory
Creating a directory when you're...

unpacking a zip archive:
```
[INFO] File 12 extracted to "/home/marcospb19/ouch/testing/.tmpmiVaZQ/src/commands/"
```
unpacking a tar (tar.gz.*) archive:
```
[INFO] "testing/.tmpcBIpss/src/commands" extracted. (0 B)
```

The output of both lines should be identical.

For this, we can have a message like `"Directory \"...\" created."`.

---

To check if a tar entry is a directory, you can call `file.header.entry_type` for a tar Entry, here:

https://github.com/ouch-org/ouch/blob/5a97029bde1d0099059e0f0e6d92a82cb9266f4a/src/archive/tar.rs#L31

and check https://docs.rs/tar/latest/tar/enum.EntryType.html#variant.Directory with [.is_dir()](https://docs.rs/tar/latest/tar/enum.EntryType.html#method.is_dir).

mvanhorn:
## Summary

Unifies directory output messages across tar and zip archive unpacking.

## Why this matters

When unpacking archives, directory entries showed confusing output:
- **zip** (`src/archive/zip.rs:56`): `"File 12 extracted to \"src/commands/\""` - directories aren't files
- **tar** (`src/archive/tar.rs:83`): `"extracted (0 B) \"src/commands\""` - the 0 B size is misleading

Both should show the same clear message indicating a directory was created.

## Changes

- `src/archive/zip.rs`: Changed directory branch to print `"Directory {path} created"` instead of `"File {idx} extracted to {path}"`
- `src/archive/tar.rs`: Added `entry.header().entry_type().is_dir()` check so directories print `"Directory {path} created"` while regular files keep the existing `"extracted ({bytes}) {path}"` format
- Updated snapshot test to match the new output

## Testing

`cargo test` passes (13/13 tests). Snapshot updated for `ui_test_ok_decompress_multiple_files`.

Fixes #328

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/src/archive/tar.rs
+++ b/src/archive/tar.rs
@@ -80,11 +80,12 @@ pub fn unpack_archive(reader: impl Read, output_folder: &Path) -> Result<u64> {
             _ => continue,
         }
 
-        info!(
-            "extracted ({}) {}",
-            BytesFmt(entry.size()),
-            PathFmt(&output_folder.join(entry.path()?)),
-        );
+        let full_path = output_folder.join(entry.path()?);
+        if entry.header().entry_type().is_dir() {
+            info!("Directory {} created", PathFmt(&full_path));
+        } else {
+            info!("extracted ({}) {}", BytesFmt(entry.size()), PathFmt(&full_path));
+        }
         files_unpacked += 1;
     }
 
--- a/src/archive/zip.rs
+++ b/src/archive/zip.rs
@@ -53,7 +53,7 @@ where
 
         match file.name().ends_with('/') {
             _is_dir @ true => {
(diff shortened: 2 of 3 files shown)


marcospb19:
Hi!

Great PR, but by a huge coincidence this change was just covered by this other PR: https://github.com/ouch-org/ouch/pull/937

That one came first so I'll prioritize that one, don't want to desincentivize improvements like this one tho, we very much appreciate these.

PR #950, opened 2026-04-19
fix multiple panics on extraction

### LINKED ISSUE #942: Suggestion: Return CLI Errors Instead of Panicking on Invalid User Input
Hi ouch maintainers, thanks for the great project！

I noticed two CLI-input-triggerable panic paths that might be better handled as normal CLI errors.

### Version
0.6.1 (And I re-checked this lines still occurs in the git repository's newest version)

### Compress case

When running `./ouch compress input1.txt input2.txt archive.bz3 --yes`, it currently panics with `output path should contain a compression format` at `src/check.rs`, and a friendlier approach would be to return a normal validation error (e.g., multi-file compression requires an archive format like `tar.*` or `zip`) instead of panicking.

In check_invalid_compression_with_non_archive_format (check.rs:266), consider replacing:
```rust
build_archive_file_suggestion(...).expect(...)
```
with graceful fallback/error construction on this user-controlled path, for example:
```rust
let suggested_output_path = build_archive_file_suggestion(output_path, ".tar").ok_or_else(|| {
    FinalError::with_title(format!("Cannot compress to '{}'.", EscapedPathDisplay::new(output_path)))
        .detail("Could not build archive suggestion from output path.")
        .hint("If compressing multiple files, try an archive format like '.tar.bz3' or '.zip'.")
})?;
```

### List case

When running `./ouch list invalid_input.bin --format tar.bz3 --yes`, it currently panics with `called Result::unwrap() on an Err value: InvalidSignature` at `src/commands/list.rs:53`, and a friendlier approach would be to propagate this as a regular CLI decoding error (invalid/corrupted bz3 stream) instead of panicking.

In list_archive_contents (commands/list.rs:57), consider replacing:
```rust
Bzip3 => Box::new(bzip3::read::Bz3Decoder::new(decoder).unwrap()),
```
with graceful fallback/error construction on this user-controlled path, for example:
```rust
Bzip3 => {
    let bz3 = bzip3::read::Bz3Decoder::new(decoder).map_err(|e| {
        crate::error::FinalError::with_title("Cannot list archive contents")
            .detail(format!("Invalid or corrupted bzip3 stream: {e}"))
    })?;
    Box::new(bz3)
},
```

Thanks for ouch again!!

curious-rabbit:
A few bugs that cause panics or silently lose info

- `ouch list` panics on a corrupt `.bz3` file because `Bz3Decoder::new` is called with `.unwrap()`. `decompress.rs` already uses `?` for the same call.
- Tar compression panics if the file walker hits an I/O or permission error mid-walk (`entry.unwrap()` in `archive/tar.rs`). Zip and 7z already use `?`.
- 7z extraction silently drops mtime-set errors via `.unwrap_or_default()`, so users get files with wrong timestamps and no warning.
- Typo in `SUPPORTED_ALIASES`: `"tzlma"` should be `"tlzma"` (every other place in the codebase uses `tlzma`).
- `ouch` panics at startup if the current directory is unreadable or has been deleted (`env::current_dir().expect(...)` in `main.rs`).
- `--threads 0` is handled inconsistently: `cli/mod.rs` treats it as "auto" but `commands/mod.rs` still passes it to rayon, and `build_global().unwrap()` panics if the thread pool is already initialized.
- `ouch compress input1.txt input2.txt archive.bz3 --yes` panics with "output path should contain a compression format" at `src/check.rs`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,10 +25,13 @@ Categories Used:
 
 ### Improvements
 
+- Report mtime-set errors for `.7z` as warnings (https://github.com/ouch-org/ouch/pull/950)
+
 ### Bug Fixes
 
-### Tweaks
+- Fix various panics not handled gracefully (https://github.com/ouch-org/ouch/pull/950)
 
+### Tweaks
 
 ## [0.7.0](https://github.com/ouch-org/ouch/releases/tag/0.7.0)
 
--- a/src/archive/sevenz.rs
+++ b/src/archive/sevenz.rs
@@ -50,13 +50,15 @@ where
                 io::copy(reader, &mut writer)?;
 
                 use filetime_creation as ft;
-                ft::set_file_handle_times(
+                // Surface mtime-set failures as warnings so users know timestamps weren't preserved
+                if let Err(e) = ft::set_file_handle_times(
                     writer.get_ref().file(),
                     Some(ft::FileTime::from_system_time(entry.access_date().into())),
                     Some(ft::FileTime::from_system_time(entry.last_modified_date().into())),
                     Some(ft::FileTime::from_system_time(entry.creation_date().into())),
-                )
-                .unwrap_or_default();
+                ) {
+                    warning!("could not set timestamps on {}: {e}", PathFmt(&file_path));
+                }
             }
 
             files_unpacked += 1;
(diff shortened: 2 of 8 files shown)


marcospb19:
Thanks!

Issue #954, opened 2026-04-21
0.7.0 GitHub release reporting old version


shyun3:
### Version

0.7.0

### Description

Release 0.7.0 is reporting the wrong version.

### Current Behavior

```sh
$ ouch --version
ouch 0.6.1
```

### Expected Behavior

```sh
$ ouch --version
ouch 0.7.0
```

### Additional Information

Downloaded from here: https://github.com/ouch-org/ouch/releases/download/0.7.0/ouch-x86_64-unknown-linux-gnu.tar.gz

marcospb19:
Thanks for reporting, I fear I'll break someones package if I try to mess with that version, so I'll just create 0.7.1.

Btw not the first time I forget to update the Cargo.toml version.

marcospb19:
Fixed in https://github.com/ouch-org/ouch/releases/tag/0.7.1

PR #957, opened 2026-04-22
Use cargo build --locked

pavelzw:
the 0.7.1 version of ouch has an outdated lockfile, see https://github.com/conda-forge/ouch-feedstock/pull/9

Would be nice to catch this early

### DIFF
--- a/.github/workflows/build-artifacts-and-run-tests.yml
+++ b/.github/workflows/build-artifacts-and-run-tests.yml
@@ -158,7 +158,7 @@ jobs:
         # there's no way to run tests for ARM64 Windows for now
         if: matrix.target != 'aarch64-pc-windows-msvc'
         run: |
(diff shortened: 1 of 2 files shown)


pavelzw:
@marcospb19 ready for review

marcospb19:
Thanks, in a quick attempt to fix #954 I edited Cargo.toml without running locally for Cargo.lock

This is truly lacking any guardrails/automations that prevent me from messing up.

marcospb19:
@pavelzw @orhun sorry for the dumb question but, is this a dealbreaker that demands a new version for your packages?

pavelzw:
Not for me, I just added a patch in conda-forge

orhun:
I just had to skip doing `--locked` which breaks the reproducibility of the package

PR #962, opened 2026-04-25
fix decompression without dir 

### LINKED ISSUE #958: Ouch 0.7.1 breaks decompression
### Version

0.7.1

### Description

running decompression in an otherwise empty directory with just the archive results in:
```
ouch d test.zip
Do you want to overwrite "."? [y/n/r/m]
```

When pressing y
```
y
[ERROR] Refusing to delete the current working directory
 - Path "." is the current directory

hint: Use a different output directory with `--dir` / `-d`
```

currently ouch only works when using the -d option

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Additional Information

_No response_

### LINKED ISSUE #963: optional smart unpack
### Version

0.7.1

### Description

After #962, running `ouch d archive.zip` without `--dir` extracts files directly into the current directory. That matches what tar, unzip, unrar, 7z and most other CLI archivers do by default, but it leaves a few annoyances that have come up before. e.g extracting an archive scatters its contents into whatever directory you happen to be in. People downloading archives into a populated directory often want a one-shot "extract this into its own folder" behavior.

The 0.6.1 smart_unpack feature tried to address this. It was removed in #907 with this rationale, quoted from the commit message:

> all this feature did was automatically flatten the top-level directory of an
> archive when decompressing an archive with a single element in its root.
> It's being removed right now for being unpredictable, might be re-added in
> the future with flags or config file, basically, something that makes it
> opt-in and not the default.

Proposal: add an opt-in flag that makes ouch extract into `CWD/<archive_basename>/` instead of directly into the CWD.

Two things that should work differently compared to the old smart_unpack:

1. The output path is always derived from the archive's filename. The user can predict where files will land just by reading the command line, without inspecting the archive first. This is the part of smart_unpack that #907
called unpredictable
2. After extraction, if the wrapper directory turns out to contain exactly one entry whose name equals the archive basename, that one entry is promoted up one level and the now-empty wrapper is removed.

The extraction can write directly into the final wrapper without temp dir, and the optional flatten can be done with a sibling staging directory and a three-step rename.

Feedback?

valoq:
Addresses #958

Sorry for the duplicate PR, git hates me today

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,42 @@
+# Dependabot configuration for ouch.
+# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    open-pull-requests-limit: 10
+    labels:
+      - "dependencies"
+    commit-message:
+      prefix: "deps"
+    cooldown:
+      default-days: 7
+    groups:
+      # Patch and minor bumps land together to reduce PR churn.
+      # Major bumps still get one PR each so they get a proper review.
+      cargo-patch-and-minor:
+        update-types:
+          - "patch"
+          - "minor"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    open-pull-requests-limit: 5
+    labels:
+      - "dependencies"
+      - "ci"
+    commit-message:
+      prefix: "ci"
+    cooldown:
+      default-days: 7
+    groups:
+      # We are less likely to be affected by breaking changes in GitHub Actions
+      # So Major bumps are accepted to reduce the amount of pull requests
+      github-actions:
+        patterns: ["*"]
--- a/.github/workflows/build-artifacts-and-run-tests.yml
+++ b/.github/workflows/build-artifacts-and-run-tests.yml
@@ -109,7 +109,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Install cross
         if: matrix.use-cross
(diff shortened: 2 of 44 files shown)


valoq:
~~I changed the decompression behavior that was added in 0.7.1 back to the one in 0.6.1 where archived decompressed in the current directory are written into a new directory with the archive name.~~

Why extracting the a new dir is better:
- it provides a way better user experience that does not accidentally mess up the cwd
- it matches user expectations from previous version
- it avoid breaking scripts that rely on the previous expectations
- is is a crucial dependency for moving forward with #723: by using an extraction directory, we can 1. create the new dir, 2. restrict write to cwd and parents and only whitelist the extraction dir, 3. extract the archive (dangerous)

Edit:
I realized that this behavior was provided by smart unpack which was removed. Going open an issue to discuss this specific issue for a future change 

~~This PR now only fixes the issue with the missing --dir option~~

Edit2: added the solution discussed in #963

valoq:
This PR now implements the decompression as discussed in https://github.com/ouch-org/ouch/issues/963

valoq:
I will add some more tests to ensure this does not overlook any edge cases

marcospb19:
Alright :+1:, when you're done maybe I'll add a test too.

valoq:
Ok, the tests are now all successful. This should be ready to be merged

marcospb19:
You also fixed `ouch decompress file.tar --dir .`

majestic...

I'll add a test for that to ensure it's not broken in the future.

marcospb19:
@valoq I pushed 2 new tests I'd like you to take a look and review in the last commit.

We can't cover all the scenarios nicely, now when `--dir` points to CWD, it'll also overwrite without asking, and it is a little inconsistent when compared to `--dir` pointing to other folders, that's not a big problem, just something I thought I'd add a test to ensure the behavior isn't changed unexpectedly.

To entirely solve this, we'd need to check for conflicts on every single file extracted, that's possible and I was working on that on another branch but it sounds simpler than it really is, there is a lot of emergent complexity for merging files on conflicts and correctly decompressing files even when a parent folder was renamed, by taking over the responsibility of conflict handling we also have to add more code to check for vulnerabilities that crates check for us.

---

Well anyways, please review the extra code and tell me if you see any problem, I think it's great for the scope of this PR, and the rest can be addressed properly later.

Signed: a929d4c4e

marcospb19:
Just did a rebase to solve conflicts.

valoq:
> Well anyways, please review the extra code and tell me if you see any problem, I think it's great for the scope of this PR, and the rest can be addressed properly later.

LGTM

valoq:
@marcospb19 should be good to merge

valoq:
The flatten logic leads to extracted files without a dir when an archive with a single file is extraced. This is a bug in my assumption that is required for the sandbox.
I will make a new patch tomorrow when I have the state of mind for it. The results I intended was that there is never an extracted archive content without first create a dir with the name of the archive so we can lock that down in landlock. The current behavior is also an inconsistency

PR #964, opened 2026-04-28
feat: improve display of overwrite choices #959

### LINKED ISSUE #959: Improve display of overwrite choices
Hi, new user here. I tried to decompress an archive and was met with this message:
```
Do you want to overwrite "."? [y/n/r/m]
```

`y` and `n` are pretty obvious, but I had no idea what `r` and `m` meant in this context. I didn't find anything in the docs or help to explain them. Instead, I had to look through the code.

Could these choices be shown as something more user friendly? Maybe something like `[(y)es/(n)o/(r)ename/(m)erge]`?

tommady:
close #959

### DIFF
--- a/src/utils/question.rs
+++ b/src/utils/question.rs
@@ -263,10 +263,8 @@ impl<'a, T: Default> ChoicePrompt<'a, T> {
             answer.make_ascii_lowercase();
             let answer = answer.trim();
 
-            if answer.is_empty() {
-                if !self.choises.is_empty() {
-                    return Ok(self.choises.remove(0).value);
-                }
+            if answer.is_empty() && !self.choises.is_empty() {
+                return Ok(self.choises.remove(0).value);
             }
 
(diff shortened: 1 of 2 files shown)


marcospb19:
A small problem that wasn't introduced in this PR:

We don't clearly state that pressing ENTER will be interpreted as `Yes`, any thoughts on how to handle this? We could have Y be uppercase and the rest lowercase, that's how people usually do it, but I'm not sure, what do you think.

tommady:
> A small problem that wasn't introduced in this PR:
> 
> We don't clearly state that pressing ENTER will be interpreted as `Yes`, any thoughts on how to handle this? We could have Y be uppercase and the rest lowercase, that's how people usually do it, but I'm not sure, what do you think.

that sounds great, the prompt would then look like this:
```bash
[(Y)es/(n)o/(r)ename/(m)erge]
```
I don't have any further suggestions to add to that!

marcospb19 (approved):
LGTM! thanks

PR #967, opened 2026-04-29
chore: migrate Zlib and Bzip2 to pure Rust implementations (#948)

### LINKED ISSUE #948: Using Rust Implementations for Zlib and Bzip2
- https://github.com/trifectatechfoundation/zlib-rs
- https://github.com/trifectatechfoundation/libbzip2-rs

tommady:
close #948

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -406,15 +406,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "cmake"
-version = "0.1.58"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c0f78a02292a74a88ac736019ab962ece0bc380e3f977bf72e376c5d78ff0678"
-dependencies = [
- "cc",
-]
-
-[[package]]
 name = "cmov"
 version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


tommady:
hey @marcospb19 
please help review while you have time.
thanks~

valoq:
Please also set 
```gzp = { version = "2.0.1", default-features = false, features = ["snappy_default"] } ```
without it gzp's defaults still pull in libz-ng-sys and libdeflater (both C), and flate2 picks zlib-ng over zlib-rs at runtime, so the C zlib stays and #948 isn't actually closed. We should also drop libbz2 and libz from the README runtime-deps list

**Comment by tommady** [comment]
> Please also set `gzp = { version = "2.0.1", default-features = false, features = ["snappy_default"] } ` without it gzp's defaults still pull in libz-ng-sys and libdeflater (both C), and flate2 picks zlib-ng over zlib-rs at runtime, so the C zlib stays and #948 isn't actually closed. We should also drop libbz2 and libz from the README runtime-deps list

Thanks for catching that! 

Additionally, I've entirely removed the "Runtime Dependencies" section from the README.md.

Since we are now using pure Rust implementations for zlib and bzip2, and bzip3 is statically linked via the bundled feature (bzip3 = { ..., features = ["bundled"] }), users will no longer need any of those C libraries installed on their systems at runtime regardless of whether they are using the musl build or not!

**Comment by marcospb19** [comment]
> users will no longer need any of those C libraries installed on their systems at runtime regardless of whether they are using the musl build or not!


That's awesome.

I checked linking info, `ldd` now only shows these two lines:

```
libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x00007fd115000000)
libm.so.6 => /usr/lib/libm.so.6 (0x00007fd116344000)
```

That are not shown for other Rust binaries here in my machine.

No more libbz2.

marcospb19:
force-pushing to solve hella conflicts

PR #969, opened 2026-04-30
update dependencies

valoq:
dependency bumps:
  brotli           7.0.0  -> 8.0.2
  bstr             1.10.0 -> 1.12.1
  bzip3            0.9.0  -> 0.12.0
  clap             4.5.20 -> 4.6.1
  flate2           1.0.30 -> 1.1.9
  fs-err           2.11.0 -> 3.3.0
  gzp              2.0.1  -> 2.0.2
  libc             0.2.155 -> 0.2.186
  lz4_flex         0.11.6 -> 0.13.0
  lzma-rust2       0.14.2 -> 0.16.2
  num_cpus         1.16.0 -> 1.17.0
  rayon            1.10.0 -> 1.12.0
  sevenz-rust2     0.19.1 -> 0.21.0
  tar              0.4.42 -> 0.4.45
  tempfile         3.10.1 -> 3.27.0
  time             0.3.36 -> 0.3.47
  unrar            0.5.7  -> 0.5.8
  zstd             0.13.2 -> 0.13.3

PR #970, opened 2026-04-30
use dependabot

valoq:
Don't merge this before #969 and #967 are merged or it will spam us with merge request


marcospb19:
We used to have dependabot but it was too spammy, maybe it'll be more bearable now only opening PRs on Monday.

PR #973, opened 2026-04-30
Make ouch builds reproducible

valoq:
This enables reproducible builds for ouch


marcospb19:
There is now a new warning in CI probably due to the addition of `rust-toolchain.toml`

https://github.com/ouch-org/ouch/actions/runs/26013866440/job/76459804293#step:9:10

I was also wondering if our CI would install Rust twice, once in the 'install rust' phase, then another when cargo is run, but looks like it's not, just the warning.

PR #974, opened 2026-05-02
feat: use sigstore signing

valoq:
This PR adds automated signature creation using sigstore.

Sigstore is a modern approach to signing source code and binary releases that does not require developers to maintain and secure private signature keys and simplify the signature process. Effectively signatures are done using short lived signing keys bound to the authentication process. Validation is done through a transparency log approach. this means all release signatures are logged in the public transparency log, allowing everyone to verify that the signature was authenticated in github. It will also list all signatures created, so even if the repo were to be compromised, all new signatures would be visible and cannot be hidden.

The advantages:
- no private (pgp) key management
- all users can verify valid releases were authenticated by github (and distro maintainers can verify the release too)
- maintainers can also check the transparency log to verify that all issued releases align with what they released.
- Fully automated on github, zero future work


marcospb19 (approved):
Pretty cool!

PR #975, opened 2026-05-02
Handle GNUSparse archive entries during tar decompression

kilianbartz:
Currently, compressing files into a tar archive via `ouch compress` may create `GNUSparse` archive entries if appropriate, since the tar crate's `Builder` is initialized with the `sparse` option set to `true`. However, during decompression, this entry type is not handled which leads to silently skipped archive entries.
Handling this entry type using `unpack_in` matches the decompression result of GNU tar for sparse archive entries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ Categories Used:
 ### Bug Fixes
 
 - Fix various panics not handled gracefully (https://github.com/ouch-org/ouch/pull/950)
+- Handle GNUSparse archive entries during tar decompression (https://github.com/ouch-org/ouch/pull/975)
 
 ### Tweaks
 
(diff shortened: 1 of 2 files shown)


tommady:
Hi @kilianbartz,
Thanks for your contribution!

If you don't mind, could you add a simple e2e test for this case?

This will help ensure that the behavior doesn't break in the future (e.g., during refactors or dependency updates).

Thank you

marcospb19 (approved):
Thanks!

(due to the delay, I think I'll let this one slide without a test :v)

kilianbartz:
Apologies for the delay, I missed GitHub's notification for @tommady's comment. I will open a new PR with the test if that's ok.

marcospb19:
haha no problem, that's OK

I think a test like this would need to create a tar archive using the 'tar' library we're already using

PR #977, opened 2026-05-06
add alias for ebooks (epub)

simonomi:
similar to `.cbz` [comic books](https://github.com/ouch-org/ouch/pull/835), [`epub`s are just zip files](https://en.wikipedia.org/wiki/EPUB), so this seems like a natural alias to add

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,9 @@ Categories Used:
 ### Tweaks
 
 - Unify directory extraction message for zip and tar archives (https://github.com/ouch-org/ouch/pull/937)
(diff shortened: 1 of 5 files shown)


marcospb19:
I messed up, then tried rebasing and GitHub closed the PR ;-; immediately.

Sorry, will try to fix.

marcospb19:
Thanks for the help!

I didn't know that about .epub, I merged it in the other PR.

EDIT: fuk, I had the co-author message in the commit of the PR but forgot about the co-author when editing the squash message, so you didn't appear in the commit itself, I'm sorry for that, I truly appreciate the contribution tho, will be more careful so I don’t make the same mistake again.

PR #980, opened 2026-05-09
build: improve man pages for subcommands

figsoda:
https://github.com/clap-rs/clap/pull/5331 added [`clap_mangen::generate_to`](https://docs.rs/clap_mangen/latest/clap_mangen/fn.generate_to.html), which automatically generates man pages for subcommands as well

this makes the generated man pages for subcommands a little nicer, showing `ouch compress` instead of `ouch-compress`


marcospb19 (approved):
Thanks!

PR #982, opened 2026-05-17
tests: added one test for decompress without --dir

venoosoo:
Adds a test for decompressing without `--dir`/`-d` by using `.current_dir()` 
instead, covering a gap noted in  https://github.com/ouch-org/ouch/issues/958#issuecomment-4298073620

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -1384,6 +1384,6 @@ fn decompress_into_cwd() {
         .arg(&compressed_archive)
         .arg("--yes")
         .assert()
-    .success();
+        .success();
     assert_same_directory(before, &after.join("before"), false);
 }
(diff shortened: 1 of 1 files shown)


marcospb19:
Hi! Thanks for the test.

Possibly due to some recent changes it's now failing with this message:

```
---- decompress_into_cwd stdout ----

thread 'decompress_into_cwd' (1864790) panicked at tests/utils.rs:72:49:
called `Result::unwrap()` on an `Err` value: Custom { kind: NotFound, error: Error { kind: ReadDir, source: Os { code: 2, kind: NotFound, message: "No such file or directory" }, path: "/tmp/.tmpaIVyCN/after/before" } }


failures:
    decompress_into_cwd
```

Could you check if that's a problem with the test or the new version of ouch? thanks!

PR #991, opened 2026-05-18
dependabot improvements

figsoda:
- enable cooldowns to reduce the risk of supply chain attacks, since it's not reasonable to audit every dependabot update
- group all github actions updates, since we are less likely to be affected by their breaking changes


marcospb19:
thx!

PR #992, opened 2026-05-18
Add Appimage release

valoq:
This adds the creation of an appimage build to the release

Uses musl libc to avoid dependency to glibc

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,4 @@
 <!--
 If your code changes text output, you might need to update snapshots
 of UI tests, read more about `insta` at CONTRIBUTING.md.
-
-Remember to edit `CHANGELOG.md` after opening the PR.
 -->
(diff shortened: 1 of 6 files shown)


marcospb19:
Hey, I haven't done any packages myself and don't intend to maintain them, if you want to add this one, it's OK! But I'd redirect any upcoming maintainance issues to you instead of dealing with them myself.

I don't know about how appimage folks organize their stuff, but I wonder if an appimage package couldn't live outside of our repo :thinking: just like all other packages, like nixpkgs and AUR and pacman.

Not a big blocker, just saying and wondering. It's your call! xd

valoq:
I haven't maintained appimage before either, but we can try that.
If there are too many problems we can just drop it again. Its a nice to have

valoq:
This seems to work just fine now and creates the appimage binary at release

ready to merge

marcospb19 (approved):
I rebased to solve conflicts and removed some stuff from CHANGELOG.md, as I'm abandoning it in favor of having an automatic changelog generated by GitHub on releases.

Dealing with that file was a big waste of time I forced on ppl.

marcospb19:
Thx again

PR #995, opened 2026-05-19
add landlock sandbox

valoq:
New version of #723

This PR adds landlock filesystem isolation to ouch as discussed in https://github.com/ouch-org/ouch/issues/722
With the new smart unpack by default feature, this is now easy to apply.

Since this is a major feature change and can easily break things, I say this should wait until after a new release has been made and then it can be merged sit in the repo for contributors to test for a while

### DIFF
--- a/src/commands/decompress.rs
+++ b/src/commands/decompress.rs
@@ -1,4 +1,6 @@
 use std::{
+    collections::HashSet,
+    ffi::OsString,
     io::{self, BufReader, Read},
     ops::ControlFlow,
     path::{Path, PathBuf},
@@ -16,7 +18,7 @@ use crate::{
     info, info_accessible,
     non_archive::lz4::MultiFrameLz4Decoder,
     utils::{
-        self, BytesFmt, LZMA_MEMLIMIT_BYTES, LimitedReader, PathFmt, copy_limited_decompression, file_size,
+        BytesFmt, LZMA_MEMLIMIT_BYTES, LimitedReader, PathFmt, copy_limited_decompression, file_size,
         io::{ReadSeek, lock_and_flush_output_stdio},
         is_path_stdin, resolve_path_conflict, user_wants_to_continue,
     },
@@ -28,8 +30,6 @@ pub struct DecompressOptions<'a> {
     /// Example: [Gz, Tar] (notice it's ordered in decompression order)
     pub formats: Vec<Extension>,
     pub output_dir: &'a Path,
-    /// Used when extracting single file formats and not archive formats
-    pub output_file_path: PathBuf,
     /// Whether the user passed `--dir` explicitly. When false (and `here` is false),
     /// archives are extracted into a basename-derived subdirectory of the CWD.
     pub output_dir_was_explicit: bool,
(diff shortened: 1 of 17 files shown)


valoq:
This PR is now pretty much ready to be tested in practice by anyone interested.

There are still have a few things I want to clean up in the unit tests but the sandbox code itself should be complete now. 
The kernel requirement is >= 6.12 and the sandbox isolates the file system not just against file manipulation but also prevents execution and network access.

valoq:
@marcospb19 This should also be ready to merge now.
Since it has still a decent chance of breaking things, I would suggest a new minor release with the recent commit before merging this into main, so it can sit there for a bit while more users test it.

marcospb19:
I messed up a merge conflict, trying to solve it right now.

marcospb19:
I'm fixing by resetting to your last good commit, and running a rebase against origin/main, that's why the force push.

src/commands/decompress.rs
```diff
@@ -97,47 +155,40 @@ pub fn decompress_file(options: DecompressOptions) -> Result<()> {
...
-                QuestionAction::Decompression,
-            )? {
-                Some(file) => file,
-                None => return Ok(()),
+            let PreparedTarget::Target {
+                dir,
+                file_name: Some(file_name),
+            } = options.prepared
+            else {
+                return Ok(());
```
marcospb19: Can you do a match to put an unreachable! here for when `file_name` is `None`? It's something that shouldn't happen but we want to notice if it happens.

src/commands/decompress.rs
```diff
@@ -179,16 +230,19 @@ pub fn decompress_file(options: DecompressOptions) -> Result<()> {
...
+            let PreparedTarget::Target { dir, .. } = options.prepared else {
+                return Ok(());
+            };
+            unpack_archive(|output_dir| unpack_fn(reader, output_dir, options.password), dir)?
         }
         #[cfg(feature = "unrar")]
         Rar => {
+            let PreparedTarget::Target { dir, .. } = options.prepared else {
+                return Ok(());
+            };
```
marcospb19: We are not checking if options.prepared is Cancelled for the 'Rar' branch arm.

Is it intentional? (If not, maybe you can put this check above this match so it's not repeated every time.)

src/commands/list.rs
```diff
@@ -25,6 +25,8 @@ pub fn list_archive_contents(
     list_options: ListOptions,
     question_policy: QuestionPolicy,
     password: Option<&[u8]>,
+    // Pre-opened tempfile FD for multi-format RAR list under the sandbox.
+    #[allow(unused_variables)] rar_spill_tempfile: Option<tempfile::NamedTempFile>,
```
marcospb19, line 29: I added this allow in a new commit (ignore its commit title).

marcospb19 (changes requested):

src/commands/decompress.rs
```diff
@@ -39,6 +39,64 @@ pub struct DecompressOptions<'a> {
...
+    question_policy: QuestionPolicy,
+    claimed_targets: &mut HashSet<PathBuf>,
+) -> Result<PreparedTarget> {
+    let (first_extension, _) = split_first_compression_format(formats);
+    let is_archive = matches!(first_extension, Tar | Zip | Rar | SevenZip);
+
+    // --dir and --here unpack into output_dir as is
+    // default mode makes a new directory named after the archive so report.txt.gz becomes report/report.txt
+    let target = if output_dir_was_explicit || here {
+        output_dir.to_path_buf()
```
marcospb19, line 94: Hey, something changed around the output file path.

Now if I run 

```sh
ouch decompress a.gz --dir out
```

It'll check for conflict on `out/`, but before it was checking against `out/a`.

So `--dir` semantics changed.

Here is a test to cover this exact behavior.

```rust
#[test]
fn decompress_single_file_dir_allows_non_empty_output_dir_with_no() {
    let (_tempdir, dir) = testdir().unwrap();

    fs::write(dir.join("a"), b"alpha").unwrap();
    crate::utils::cargo_bin()
        .args(["compress", "a", "a.gz", "--yes"])
        .current_dir(dir)
        .assert()
        .success();
    fs::remove_file(dir.join("a")).unwrap();

    fs::create_dir(dir.join("out")).unwrap();
    fs::write(dir.join("out").join("other-file"), b"keep").unwrap();

    crate::utils::cargo_bin()
        .args(["decompress", "a.gz", "--dir", "out", "--no"])
        .current_dir(dir)
        .assert()
        .success();

    assert_eq!("alpha", fs::read_to_string(dir.join("out").join("a")).unwrap());
    assert_eq!("keep", fs::read_to_string(dir.join("out").join("other-file")).unwrap());
}
```

It wouldn't fail before (I checked), but fails now on this branch.

Is it intended?

src/commands/list.rs
marcospb19, line 29: AI reviewer detected that we now leave a dangling temporary folder.

To confirm it:

```sh
before=$(find /tmp -maxdepth 1 -type d -name '.tmp*' | sort)
cargo run -- list tests/data/testfile.rar3.rar.gz > /dev/null
after=$(find /tmp -maxdepth 1 -type d -name '.tmp*' | sort)
echo diff: && comm -13 <(printf "%s\n" "$before") <(printf "%s\n" "$after")
```

It means that every time it's run we add one more folder at /tmp.

Maybe the issue is we have permission to delete files inside the folder but not the folder itself? (would require parent permission).

src/commands/decompress.rs
valoq: I moved the cancelled check above the match so every arm including rar shares it

src/commands/decompress.rs
valoq: No. The conflict now checks the output file out/a instead of the directory, so a non-empty out/ no longer blocks it. also added the test

src/commands/list.rs
valoq: Should work now by allowing only the deletion of empty directories in /tmp.

marcospb19:
> I would suggest a new minor release with the recent commit before merging this into main

done, 0.8.1 was released, will review

marcospb19 (approved):
Thanks for the review follow ups, looks good now.

It's been almost 2 years since the first landlock draft, a lot changed to make this happen.

Thanks for doing it all 🏆.

PR #996, opened 2026-05-20
fix new extract logic

valoq:
This PR ensures that extraction always happens inside the new dir that is created before the archive is parsed. This is a requirement for future sandbox implementations and it fixes the behavior that is intended to always produce a directory with the archive name. Previously archives with single files would be flatten and end up with files in the $CWD.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Categories Used:
 
 ### New Features
 
-- Unpack into new folder by default (https://github.com/ouch-org/ouch/pull/962).
+- Unpack into new folder by default (https://github.com/ouch-org/ouch/pull/962) and (https://github.com/ouch-org/ouch/pull/996).
 - Add `--here` flag to unpack into current directory (https://github.com/ouch-org/ouch/pull/962).
 - List: show symlink targets (for tar and zip) (https://github.com/ouch-org/ouch/pull/934)
 - Add aliases for ebooks (`.epub`) (https://github.com/ouch-org/ouch/pull/981)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,11 +15,23 @@ Categories Used:
 
 **Bullet points in chronological order by PR**
 
-## [Unreleased](https://github.com/ouch-org/ouch/compare/0.7.1...HEAD)
+## [Unreleased](https://github.com/ouch-org/ouch/compare/0.8.0...HEAD)
 
 ### New Features
 
-- Unpack into new folder by default (https://github.com/ouch-org/ouch/pull/962) and (https://github.com/ouch-org/ouch/pull/996).
+### Improvements
+
+### Bug Fixes
+
+### Tweaks
+
+
+## [0.8.0](https://github.com/ouch-org/ouch/compare/0.7.1...0.8.0)
+
+### New Features
+
+- Releases: add Apple Silicon macOS (`aarch64-apple-darwin`) to assets.
+- Unpack into new folder by default (https://github.com/ouch-org/ouch/pull/962).
 - Add `--here` flag to unpack into current directory (https://github.com/ouch-org/ouch/pull/962).
 - List: show symlink targets (for tar and zip) (https://github.com/ouch-org/ouch/pull/934)
 - Add aliases for ebooks (`.epub`) (https://github.com/ouch-org/ouch/pull/981)
--- a/src/commands/decompress.rs
+++ b/src/commands/decompress.rs
@@ -278,8 +278,6 @@ fn unpack_archive(
 /// Expects `wrapper` to be a just-decompressed archive output directory, if
 /// `wrapper` contains exactly one entry with the same name (e.g. extracting
 /// `archive.zip` produced `archive/archive/...`), then flatten it to `archive/...`.
-///
-/// Returns the resulting path the user should see. If reads fail,
 fn deduplicate_basename_wrapper(wrapper: &Path) -> Result<()> {
     let Some(wrapper_name) = wrapper.file_name() else {
         return Ok(());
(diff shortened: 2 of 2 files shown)


marcospb19:
I love that non-archives are flattened by default, so, I'll investigate further into landlock to see if it's possible to workaround this, if not we are merging.

I'd like a workaround like:
1. Prohibit writing to current folder.
2. Allow writing only to the new file.

That'd be slightly more complex since our Landlock implementation would need to be aware of whether or not we are about to decompress a single-file format (non-archive), and also, know that path.

But if we achieve that we don't have to tweak this behavior.

valoq:
I think you are right that this can be avoided. I will need to check again if its worth it but it also depends on the archive format;
For single-stream formats (gz, bz2, xz, zst, lz4, etc.) where the output is one file with a name derived purely from the input filename we can open a FD before parsing the archive, then restrict the runtime and then do the dangerous extraction.
For multi-file archives (tar/zip/7z/rar) this cannot work though sicne we have to parse the contents of the archives before we know the target filename/structure.

Can you tell me what use case you would like to see supported?

marcospb19:
> For single-stream formats (gz, bz2, xz, zst, lz4, etc.) where the output is one file with a name derived purely from the input filename we can open a FD before parsing the archive, then restrict the runtime and then do the dangerous extraction.

Yeah, when I mentioned a workaround, I was strictly thinking about this scenario.

But,

> ... Previously archives with single files would be flatten and end up with files in the $CWD.

This :arrow_up: is usually not the case anymore.

Please correct me if I'm wrong, but in the latest `main` and since `0.8.0` there is no automatic flattening by default anymore.

Decompression will happen inside a new directory unless one of these scenarios happen:

1. `ouch d archive.zip` and `archive` already exists and user chose `merge`, will reuse that folder (can we sandbox that folder? Ouch needs to write to it, so it can also read existing files?).
2. `ouch d archive.zip --dir .` or `ouch d archive.zip --here` will do it all inside of $CWD, so we can't sandbox $CWD?
3. Non-archive formats like we discussed.

In summary, when decompression in $CWD happens:

1. User chose `merge`
2. User explicitly asked to decompress to $CWD (just works now like merge cause we don't ask to delete $CWD).
3. Non-archives are decompressed in $CWD.

And, I believe this PR doesn't tackle these 3 because it was based on an assumption for something that changed recently, can you confirm if that's right?

ofek:
Here's a Nushell script I used to confirm that the latest version already keeps single-file archives under a new directory:

<details>

```nu
let work = ($env.PWD | path join "ouch-pr-996-behavior")
rm -rf $work
mkdir $work
cd $work

git clone https://github.com/ouch-org/ouch.git repo
cd repo
git fetch origin main "+pull/996/head:refs/heads/pr-996"

let exe = if $nu.os-info.name == "windows" { "ouch.exe" } else { "ouch" }

mkdir ../bin

git switch --detach origin/main
cargo build --locked
cp (["target", "debug", $exe] | path join) (["..", "bin", $"ouch-main-($exe)"] | path join)

git switch --detach pr-996
cargo build --locked
cp (["target", "debug", $exe] | path join) (["..", "bin", $"ouch-pr996-($exe)"] | path join)

cd ..

let main_bin = ([$work, "bin", $"ouch-main-($exe)"] | path join)
let pr_bin = ([$work, "bin", $"ouch-pr996-($exe)"] | path join)

mkdir fixtures/src
"single contents" | save -f fixtures/src/single.txt
"same basename contents" | save -f fixtures/src/same
"plain gzip contents" | save -f fixtures/src/plain.txt
mkdir fixtures/src/wrapped
"wrapped contents" | save -f fixtures/src/wrapped/file.txt

^$main_bin --yes c fixtures/src/single.txt fixtures/single.zip
^$main_bin --yes c fixtures/src/same fixtures/same.zip
^$main_bin --yes c fixtures/src/wrapped fixtures/wrapped.zip
^$main_bin --yes c fixtures/src/plain.txt fixtures/plain.txt.gz

def snapshot [dir] {
  let prev = $env.PWD
  cd $dir
  let rows = (
    ls -a **/*
    | each { |it|
      let name = ($it.name | str replace -a "\\" "/")
      if $it.type == dir { $"($name)/" } else { $name }
    }
    | sort
  )
  cd $prev
  $rows
}

def run-one [bin branch case] {
  let dir = (["runs", $branch, $case.name] | path join)
  rm -rf $dir
  mkdir $dir
  cp $case.archive $dir

  if $case.name == "merge-existing-output-dir" {
    mkdir ($dir | path join "single")
    "pre-existing" | save -f (($dir | path join "single" | path join "marker.txt"))
  }

  let prev = $env.PWD
  cd $dir
  ^$bin --yes d ($case.archive | path basename) ...$case.args
  rm -f ($case.archive | path basename)
  cd $prev

  { branch: $branch, case: $case.name, tree: (snapshot $dir | str join ", ") }
}

def run-all [bin branch] {
  let cases = [
    { name: "default-single-file-archive", archive: "fixtures/single.zip", args: [] }
    { name: "default-same-basename-file", archive: "fixtures/same.zip", args: [] }
    { name: "default-duplicate-root-dir", archive: "fixtures/wrapped.zip", args: [] }
    { name: "merge-existing-output-dir", archive: "fixtures/single.zip", args: [] }
    { name: "explicit-dir-dot", archive: "fixtures/single.zip", args: ["--dir", "."] }
    { name: "explicit-here", archive: "fixtures/single.zip", args: ["--here"] }
    { name: "non-archive-gzip", archive: "fixtures/plain.txt.gz", args: [] }
  ]

  $cases | each { |case| run-one $bin $branch $case }
}

(run-all $main_bin "main") ++ (run-all $pr_bin "pr-996") | table -e
```
</details>

The output:

```
╭────┬────────┬─────────────────────────────┬───────────────────────────────────────────────╮
│  # │ branch │            case             │                     tree                      │
├────┼────────┼─────────────────────────────┼───────────────────────────────────────────────┤
│  0 │ main   │ default-single-file-archive │ single/, single/single.txt                    │
│  1 │ main   │ default-same-basename-file  │ same                                          │
│  2 │ main   │ default-duplicate-root-dir  │ wrapped/, wrapped/file.txt                    │
│  3 │ main   │ merge-existing-output-dir   │ single/, single/marker.txt, single/single.txt │
│  4 │ main   │ explicit-dir-dot            │ single.txt                                    │
│  5 │ main   │ explicit-here               │ single.txt                                    │
│  6 │ main   │ non-archive-gzip            │ plain.txt                                     │
│  7 │ pr-996 │ default-single-file-archive │ single/, single/single.txt                    │
│  8 │ pr-996 │ default-same-basename-file  │ same/, same/same                              │
│  9 │ pr-996 │ default-duplicate-root-dir  │ wrapped/, wrapped/file.txt                    │
│ 10 │ pr-996 │ merge-existing-output-dir   │ single/, single/marker.txt, single/single.txt │
│ 11 │ pr-996 │ explicit-dir-dot            │ single.txt                                    │
│ 12 │ pr-996 │ explicit-here               │ single.txt                                    │
│ 13 │ pr-996 │ non-archive-gzip            │ plain.txt                                     │
╰────┴────────┴─────────────────────────────┴───────────────────────────────────────────────╯
```

valoq:
Sorry I wasn't clear here, (and I need to find some time to get my head into this again)
Aside from the one file case, the PR isn't about where files end up, its about where files are written in the flatten process.
Before this PR, ouch 0.8.0 writes to $CWD during the flatten operation which contradicts the condition for the sandbox.
This PR keeps the flatten operation inside the target directory where write is permitted.

The question about where files should end up is a separate issue. We can probably support this, but Im unsure if we should because it depends on the archive format and adds complexity. Mayby im overthinking it though.

Edit:
I think it comes down to this: we can support the single file output but not by flattening but by extracting to a single file directly. This is only possible in the sandbox when we know the archive is a single file before the archive is extracted. Which is only possible for gz, bz2, xz, zst, lz4 - the sandbox woudl then opena  file descripor fior the target file with write permissions before the sandbox is in effect and before the archive is parsed.
But all other formats need to be parsed before we know the contents, which violates the threat model of the sandbox and makes this impossible.

So either: 
1. we support this only for the formats that allow this (gz, bz2, xz, zst, lz4)
2. we keep it simple and ensure all formats behave the same and end up in a new directory

Landlock supports both from what I can tell so I will leave the decision up to you @marcospb19

marcospb19:
@valoq thank you for correcting me and explaining in more detail.

> So either:
>
> 1. we support this only for the formats that allow this (gz, bz2, xz, zst, lz4)
> 2. we keep it simple and ensure all formats behave the same and end up in a new directory

I was indecisive about this, but, let's say that, at least for now, it's a great idea to keep it simple even if it means we won't be flattening single-file **non-archive** formats (xz, bz, bz3, lzma, zst, etc).

For these formats we can later follow up with a file descriptor workaround so we write it to $CWD instead.

While I do think it could be done in the landlock PR, if we break this into multiple steps maybe we make it easier to ourselves to get this out earlier.

TLDR: let's do the second option! We can improve the behavior later when landlock is merged (FD workaround).

marcospb19 (changes requested):
Can you solve the new conflicts on this one, I think it's good to merge after that.

marcospb19 (approved):
thanks, lgtm

marcospb19:
fyi, when squashing I changed the commit message to better convey what this is, I think I got it right this time :)

PR #999, opened 2026-05-27
fix: disambiguate archive entries when multiple inputs share the same basename

### LINKED ISSUE #978: Handle better multiple paths with the same basename
Hi

If I try to compress:

C:\scripts
E:\Backup\Scripts
E:\Local\Scripts

ouch strips path, and I end up with only "Scripts" in the archive.

In Zip, it is catastrophic, everything is inside a single folder.
In Tar, each folder is shown, but if I decompress, everything is in the same folder.

Not sure how it can be handled.

Maybe a --full-path parameter.

Could be useful (example) for:
/etc
/opt/solr/config
/usr/local/etc
/var/lib/xxx/config
/srv/solr

toller892:
## Problem

When compressing multiple directories with the same basename (e.g., `/mnt/d/Temp/GOG/Scripts` and `/mnt/d/Temp/kodi/Scripts`), ouch strips paths to just the basename using `file_name()`. This causes:

- **Zip**: `"Invalid zip archive - Duplicate filename: Scripts/"`
- **Tar**: both directories merge into one, causing data loss on decompress

Fixes #978

## Root Cause

In all three archive builders (tar, zip, 7z), the `build_archive` function:
1. Calls `cd_into_same_dir_as(explicit_path)` to change cwd to the parent directory
2. Uses `explicit_path.file_name().unwrap()` to get just the basename
3. The walker walks from the basename, so archive entries are just `Scripts/...`

When two inputs share the same basename, they collide.

## Fix

When input paths have duplicate basenames, compute the **common ancestor directory** of all inputs and store archive entries relative to it, including enough parent path components to disambiguate (e.g., `parent1/Scripts` vs `parent2/Scripts`).

When basenames don't collide, behavior is **unchanged** (backward compatible).

### Example

```
# Before (broken):
ouch compress /a/parent1/Scripts /a/parent2/Scripts archive.tar
# Archive contains: Scripts/a.txt, Scripts/b.txt  <-- COLLISION!

# After (fixed):  
ouch compress /a/parent1/Scripts /a/parent2/Scripts archive.tar
# Archive contains: parent1/Scripts/a.txt, parent2/Scripts/b.txt  <-- no collision
```

## Files Changed

- `src/utils/fs.rs` - Added `has_duplicate_basenames()` and `common_ancestor()` helper functions
- `src/archive/tar.rs` - Use common ancestor approach when basenames collide
- `src/archive/zip.rs` - Same fix for zip builder
- `src/archive/sevenz.rs` - Same fix for 7z builder
- `tests/integration.rs` - Added tests for tar and zip with duplicate basenames

## Test Results

All 91 tests pass (34 unit + 39 integration + 3 mime + 13 UI + 2 utils), including the 2 new tests:
- `compress_dirs_with_same_basename_tar`
- `compress_dirs_with_same_basename_zip`

### DIFF
--- a/src/archive/sevenz.rs
+++ b/src/archive/sevenz.rs
@@ -18,7 +18,8 @@ use crate::{
     info,
     list::{FileInArchive, ListFileType},
     utils::{
-        BytesFmt, FileVisibilityPolicy, PathFmt, cd_into_same_dir_as, ensure_parent_dir_exists, is_same_file_as_output,
+        BytesFmt, FileVisibilityPolicy, PathFmt, cd_into_same_dir_as, common_ancestor, ensure_parent_dir_exists,
+        has_duplicate_basenames, is_same_file_as_output,
     },
     warning,
 };
@@ -134,14 +135,35 @@ where
     let mut writer = sevenz_rust2::ArchiveWriter::new(writer)?;
     let output_handle = Handle::from_path(output_path);
 
+    // When multiple inputs share the same basename, archive entry names would collide.
+    // In that case, use the common ancestor of all paths and store entries relative to it,
+    // including enough parent path components to disambiguate.
+    let use_common_ancestor = files.len() > 1 && has_duplicate_basenames(files);
+    let common_ancestor = if use_common_ancestor {
+        Some(common_ancestor(files))
+    } else {
+        None
+    };
+
     for filename in files {
-        let previous_location = cd_into_same_dir_as(filename)?;
+        let previous_location = if let Some(ref ancestor) = common_ancestor {
+            let previous = env::current_dir()?;
+            env::set_current_dir(ancestor)?;
+            previous
+        } else {
+            cd_into_same_dir_as(filename)?
+        };
 
         // Unwrap safety:
         //   paths should be canonicalized by now, and the root directory rejected.
-        let filename = filename.file_name().unwrap();
+        // When using common ancestor, include enough parent path to avoid collisions.
+        let walker_root = if let Some(ref ancestor) = common_ancestor {
+            filename.strip_prefix(ancestor).unwrap().to_path_buf()
+        } else {
+            PathBuf::from(filename.file_name().unwrap())
+        };
 
-        for entry in file_visibility_policy.build_walker(filename) {
+        for entry in file_visibility_policy.build_walker(&walker_root) {
             let entry = entry?;
             let path = entry.path();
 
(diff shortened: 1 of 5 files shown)


Mazwak:
My 2 cents.

I would behave like tar/zip by default: Use path from command line, stripping leading /.

And maybe add something like --shorten-path.

marcospb19:
> I would behave like tar/zip by default: Use path from command line, stripping leading /.

I'm slightly more favorable to this suggestion than the biggest common ancestor approach used in this PR.

Mainly due to predictability.

@tommady @valoq can I get u guys opinions?

(If we change the default behavior we wouldn't need a flag too.)

tommady:
> > I would behave like tar/zip by default: Use path from command line, stripping leading /.
> 
> 
> 
> I'm slightly more favorable to this suggestion than the biggest common ancestor approach used in this PR.
> 
> 
> 
> Mainly due to predictability.
> 
> 
> 
> @tommady @valoq can I get u guys opinions?
> 
> 
> 
> (If we change the default behavior we wouldn't need a flag too.)

As I mentioned in the issue comment below, I will follow your advice.

https://github.com/ouch-org/ouch/issues/978#issuecomment-4553789711

valoq:
The tar and zip approach is better, yes.

A few more potential issues:

- Cross-drive inputs (e.g. C:\scripts and E:\Backup\Scripts) share no component, so common_ancestor returns an empty path and set_current_dir("") fails on Windows.
- Passing the same path twice now fails. The shared folder it finds is the whole path, so there is nothing left below it to add to the archive, and it errors. Before this change, the same path twice just produced a duplicate entry.
- When inputs share only the root, entries keep their full path from the root  instead of only the part needed to tell the files apart
- The 7z change needs unit tests for all cases

Mazwak:
Windows cross drive is a specific need. But it is not necessary for everything.

I tested tar and zip, and they behave differently.

tar does not support windows drive, so it could either fail, mandatory if 2 different drives, or warn it will strip drive.

zip works like ouch, it strips everything but the last directory.
(Thought it was like tar, but it isn’t)

BUT

Compress-archive (powershell) will just merge everything.
7z will refuse to create the archive with duplicate entries.

I did not test other zip tools.

7z format is like zip.

And it seems zip does not support drives anyway.


So, i think it’s a choice between

1)  just behave like original.

2) Add a fake directory for the drive

C/Scripts/…
E/Scripts/…


That’s your call, you are the dev.

PR #1008, opened 2026-06-11
Add archive dir decompress

senylar:
## Summary

Adds a new `--archive-dir` / `-a` option for `ouch decompress`.

When decompressing an archive without `--dir`, this extracts the archive next to the input archive into a basename-derived directory. When combined with `--dir`, it extracts into a basename-derived directory under the chosen output directory.

## Examples

```sh
ouch decompress summer_vacation.zip -a
ouch decompress summer_vacation.zip -a --dir pictures

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ Categories Used:
 
 ### New Features
 
+- Add `--archive-dir` / `-a` option to decompress archives into a directory named after the archive.
+
 ### Improvements
 
 ### Bug Fixes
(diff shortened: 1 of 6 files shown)


valoq:
> Extract each archive into a directory named after the archive

I'm confused. This is already how the default decompress behaves, isn't it?

marcospb19:
Hi, sorry.

Yeah I do believe this is the defualt behavior, maybe you meant something else, if so let me know.

I'm going to close this one, to ensure your PR is not closed due to some misalignment, like in this case, your safest bet is creating an issue before to discuss the solution.

PR #1011, opened 2026-06-20
fix exit code in non-interactive mode

valoq:
When `ouch decompress` hits a file or directory conflict and stdin is /dev/null e.g. when used in a script, it cannot prompt the user.
The default of FileConflitOperation is Cancel, which makes the entry silently skip and the whole run report success (exit 0) while extracting nothing.

This is also inconsistent with an empty-pipe (EOF) stdin, which already returns a FinalError and exits non-zero for the very same prompt.

This PR makes the behavior consistent and returns an error when the conflict is detected while running non interactively

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ Categories Used:
 
 ### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


marcospb19 (approved):
looking great

PR #1007, opened 2026-06-08
fix: ignore invalid unix permissions and setuid bits from zip

tommady:
Fixes #1006 

Some zip files created on MS-DOS environments store invalid attributes that `zip` reads as unix modes missing the file type bits, resulting in unintended setuid permissions. 

In the reported issue, `stat` revealed the file access mode was `0o4032`. 
In a valid Unix permission mode, the file type bits (`S_IFMT`, e.g., `0o100000` for a regular file) must be set.

A mode of `0o4032` entirely lacks these file type bits, meaning it is garbage data (likely MS-DOS attributes or arbitrary metadata from the archive generator) being mistakenly interpreted as a Unix mode.

PR #1014, opened 2026-06-28
validate zip unix permissions at file creation

valoq:
- Move the zip mode check into one helper used at both file creation and chmod
- A mode without unix file-type bits is treated as invalid and falls back to 0o644
- Fixes a gap left by #1007, where file creation still applied the bogus mode
- Drop the redundant 0o777 mask already done by sanitize_archive_mode
- Add a test for a mode with no file-type bits falling back to the default
- Add a test confirming setuid, setgid, and sticky bits are stripped


marcospb19:
thanks!

PR #1018, opened 2026-07-03
fix zip compression

valoq:
ouch writes zip files whose folder entries have an invalid size, so strict readers like 7-Zip refuse to open the archive. This started with the recent zip crate update from version 6 to 8, which no longer sets the size correctly on empty folder entries when ZIP64 is enabled. Any zip that contains a folder is affected while single files are fine.

### DIFF
--- a/src/archive/zip.rs
+++ b/src/archive/zip.rs
@@ -260,7 +260,8 @@ where
                     io::copy(&mut file, &mut writer)?;
                 }
                 FileType::Directory => {
-                    writer.add_directory(entry_name, default_options)?;
+                    // Directory entries have no data and get an invalid size when ZIP64 is forced on
+                    writer.add_directory(entry_name, default_options.large_file(false))?;
                 }
                 FileType::Symlink => {
                     let target_path = path.read_link()?;


marcospb19 (approved):

Gallardo994:
Even though it's already merged, I'd like to point out it also affects Archive Utility on MacOS, as this fix is not in a release build yet, so that whoever looks for a solution finds it.

Archive Utility produces this error:
`Unable to expand "zipname.zip" into "targetdir". (Error 79 - Inappropriate file type or format.)`

marcospb19:
@valoq I could cherry-pick this to a new branch 0.8.2 based on 0.8.1 and publish it as a hotfix without also releasing Landlock support.

Another option is we just risk releasing it with Landlock.

marcospb19:
@valoq I see you gave it a thumbs up, is that for cherry-picking this? (and not the second alternative I mentioned)

valoq:
> is that for cherry-picking this?

yes

marcospb19:
Done, released in 0.8.2.

PR #1020, opened 2026-07-04
fix size-limit and non-UTF-8 zip symlinks

valoq:
- LimitedReader: accept output of exactly OUCH_MAX_DECOMPRESSED_BYTES (not below)
- zip: read symlink targets as bytes, not UTF-8


marcospb19:
I thought zip only supported UTF-8, I'm testing locally with `zip` binary.

marcospb19 (approved):
Good catch.

PR #1026, opened 2026-07-19
Improve signing with cosign instead of github cli

valoq:
Use cosign for signature verification instead of github cli


marcospb19:
Hey, is there some piece of text you can point me to so I can better understand why this is better.

valoq:
> Hey, is there some piece of text you can point me to so I can better understand why this is better.

I had used the github cli since it was the first documented solution to add to the github ci, but 'gh' requires user authentication. I dont think thats appropriate to verify the signature.
sigstore has no requirement, is much more independent and also more common.

marcospb19:
Thanks!

PR #1030, opened 2026-07-23
Scope visibility flags to compression

### LINKED ISSUE #327: Reduce the scope of some flags
Flags like `--hidden` and `--gitignore` are set in the `Opts` struct,
however, they are specific to the `compress` subcommand, so these
flags should be moved to the `Subcommand::Compress` enum variant.

https://github.com/ouch-org/ouch/blob/5a97029bde1d0099059e0f0e6d92a82cb9266f4a/src/opts.rs#L28-L34

This will hide those flags when using `Ouch` with other commands.

kx0101:
## Summary
- move `--hidden` and `--gitignore` to the `compress` subcommand
- reject these flags for `decompress`, `list`, and the top-level command
- update CLI parsing tests and help snapshots

## Testing
- `cargo test`
- `cargo test --no-default-features`
- `cargo clippy --all-targets --all-features -- -D warnings`

Closes #327

### DIFF
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -26,17 +26,9 @@ pub struct CliArgs {
     #[arg(short = 'A', long, env = "ACCESSIBLE", global = true)]
     pub accessible: bool,
 
-    /// Ignore hidden files
-    #[arg(short = 'H', long, global = true)]
-    pub hidden: bool,
-
     /// Silence output
     #[arg(short, long, global = true)]
     pub quiet: bool,
-
-    /// Ignore files matched by git's ignore files
-    #[arg(short, long, global = true)]
-    pub gitignore: bool,
 
     /// Specify the format of the archive
     #[arg(short, long, global = true)]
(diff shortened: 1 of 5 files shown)


valoq:
LGTM, only a changelog entry is missing

valoq (approved):

kx0101:
@valoq thanks for checking it out, I just added the changelog entry 😄

marcospb19:
@valoq in 0.8 I abandoned the CHANGELOG in favor of having an exhaustive list generated by GitHub when making the releases.

The problem was that the CHANGELOG was always missing entries, and other projects auto-generate the CHANGELOG with scripts based on PRs.

For example, Zed auto-generates from a nicely formatted PR description, if it's malformed the CI complains with an error.

I know giving up the CHANGELOG gave me a sunken cost fallacy feel, and I should've consulted you before doing that :/, sorry.

But now that it's done I think we need to revert the CHANGELOG change before merging this.

kx0101:
@marcospb19 all good don't worry about it! happens 😄 

I deleted the commit

marcospb19:
Thanks!

marcospb19:
wait I was about to merge but I see some commits that aren't really part of your PR, is that by accident :thinking: or did I do something wrong here

maybe this wasn't based on main?

kx0101:
@marcospb19 the actual change was only one commit, but the branch picked up unrelated history from my fork. 

I'll rebase/cherry-pick it onto the current upstream main and force-push a clean branch

marcospb19:
Thanks!

PR #1031, opened 2026-07-23
fix: do not overwrite during merge

### LINKED ISSUE #1004: Decompression: When merging folders, files are overwritten silently
### Problem

When decompressing archives into the same destination (and merging folders), files with identical names can be silently overwritten.

This happens in the following scenarios:

1. Decompress, and on conflict, user chooses the "merge" option.
2. Decompress with `--dir .`
3. Decompress with `--here` (equivalent to --dir)

Currently, conflict detection only operates at the folder level. Once a folder has been "accepted", each archive backend writes to files independently, possibly overwriting silently.

### Expected behavior

- Check and prompt for conflicts for every single file.
- Ensure consistent behavior across all supported archive formats.

### Context

This has been a known limitation that requires a refactor. Solving it would also require moving format-specific safety/sanity checks into the shared layer, since the formats would no longer be writing files directly. And allowing for "renaming" of nested folders that will have files inside of them requires some complex routing logic. Also ensure that logs for subsequent files that had their parents renamed have their paths fixed for their new location (requires a path trie for sure).

_Originally posted by @roland-5 in https://github.com/ouch-org/ouch/discussions/998#discussioncomment-17118245_

valoq:
Fixes #1004 

Conflicts were only resolved at the folder level. After accepting a merge, or when using --dir . or --here, each backend wrote files straight to disk and replaced same named files without asking.

Now they ask first. One shared helper in utils/fs.rs is called by tar, zip and 7z before writing a regular file, so all three behave the same: rename, overwrite or skip when prompted, overwrite with --yes, skip with --no, and no prompt when nothing clashes. Only existing files count as conflicts, so directories still merge and nothing is deleted.

~~Rar is left as is. Returning false from its Start callback cancels the whole extraction instead of skipping one file, and the per entry API would drop the large dictionary guard. I have not found a solution for it yet.~~

Added a test covering skip and overwrite for each archive format. Two old tests asserted the silent overwrite and now answer the prompt.

### DIFF
--- a/src/archive/tar.rs
+++ b/src/archive/tar.rs
@@ -13,21 +13,21 @@ use fs_err as fs;
 use same_file::Handle;
 
 use crate::{
-    Result,
+    QuestionPolicy, Result,
     error::FinalError,
     info,
     list::{FileInArchive, ListFileType},
     utils::{
         self, BytesFmt, FileType, FileVisibilityPolicy, PathFmt, canonicalize, create_symlink, is_same_file_as_output,
-        read_file_type, sanitize_archive_mode, set_permission_mode, validate_dest_inside_root, validate_entry_path,
-        validate_symlink_target,
+        read_file_type, resolve_extraction_conflict, sanitize_archive_mode, set_permission_mode,
+        validate_dest_inside_root, validate_entry_path, validate_symlink_target,
     },
     warning,
 };
 
 /// Unpacks the archive given by `archive` into the folder given by `into`.
 /// Assumes that output_folder is empty
-pub fn unpack_archive(reader: impl Read, output_folder: &Path) -> Result<u64> {
+pub fn unpack_archive(reader: impl Read, output_folder: &Path, question_policy: QuestionPolicy) -> Result<u64> {
     let mut archive = tar::Archive::new(reader);
 
     let mut files_unpacked = 0;
@@ -35,6 +35,9 @@ pub fn unpack_archive(reader: impl Read, output_folder: &Path) -> Result<u64> {
 
     for entry in archive.entries()? {
         let mut entry = entry?;
+
+        // Set when the user renamed a file so the log can show the real path.
+        let mut written = None;
 
         match entry.header().entry_type() {
             tar::EntryType::Symlink => {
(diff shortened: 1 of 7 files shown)


valoq:
Rar is covered as well now

should be good for review

marcospb19:
Hi, sorry for the big delay again, I should be revisiting and reviewing stuff soon.

marcospb19:
Hey. This is addressing what I consider to be the most important problem at the moment.

I tried to solve this a couple months back but paused to prioritize the newer versions and landlock support.

--- 

One worry with this approach is, if the user chooses to rename a directory, all its children need their path to be renamed too, regardless of the children file type, so we need a path trie (data structure).

Please correct me if I'm wrong!! But below are my thoughts based on this fact.

And if we're achiving this level of control over decompression, then we need to put all path resolution behind this one magic (conflict resolution) struct that will handle it for all archives.

For this magic struct to control where things go, we need to check if all the libraries give us the control we need over decompression, I still haven't checked this thoughtfully.

However, a crazy simpler alternative to making this work for every archive code is to, again, have all decompression happen to a temporary directory folder, and then 1 generic code to move the results to the final directory doing all of this path-trie conflict resolution. There should be usability implications we must consider.

Any thoughts?

valoq:
> One worry with this approach is, if the user chooses to rename a directory, all its children need their path to be renamed too, regardless of the children file type, so we need a path trie (data structure).

this PR only prompts for regular files and directories always merge and are never renamed. Im not sure there is any rename branch that would be relevant here


> For this magic struct to control where things go, we need to check if all the libraries give us the control we need over decompression, I still haven't checked this thoughtfully.

tar and zip allow per-entry destinations, 7z mostly, rar doesnt.

> However, a crazy simpler alternative to making this work for every archive code is to, again, have all decompression happen to a temporary directory folder, and then 1 generic code to move the results to the final directory doing all of this path-trie conflict resolution. There should be usability implications we must consider.

This is only needed for rar, which is already implemented here

marcospb19:
> this PR only prompts for regular files and directories always merge and are never renamed. Im not sure there is any rename branch that would be relevant here

Makes sense, if that's the case you can disregard the rest of my comment.

src/archive/tar.rs
```diff
@@ -66,7 +69,20 @@ pub fn unpack_archive(reader: impl Read, output_folder: &Path) -> Result<u64> {
...
+                let full_path = output_folder.join(&safe_relpath);
+
+                let Some(dest) = resolve_extraction_conflict(&full_path, question_policy)? else {
+                    continue;
+                };
+
+                if dest == full_path {
+                    entry.unpack_in(output_folder)?;
+                } else {
+                    entry.unpack(&dest)?;
```
marcospb19, line 83: https://docs.rs/tar/latest/tar/struct.Entry.html#method.unpack says that this function won't do some of the sanity checks.

But Landlock will protect us.

marcospb19:
Thanks again, your approach was clear and straight forward.

PR #1033, opened 2026-07-27
fix ZIP timestamps across time zones

### LINKED ISSUE #902: Files from extracted zip have wrong modified time
### Version

0.6.1

### Description

Files extracted from a zip have the modified time updated to a value 6 hours in the past. I'm in UTC-6, so the time being read from the extracted file seems to be interpreted as UTC time rather than local time.

Attached are an empty file (`test.orig.txt`), an archive with that file (`test.zip`), the extracted result from `ouch` (`test.ouch.txt`) and the extracted result from `unzip`, which has the correct modified time (`test.unzip.txt`).

### Current Behavior

Here's the relevant `stat` output:

```
> stat test.orig.txt
  File: test.orig.txt
  Size: 0               Blocks: 0          IO Block: 4096   regular empty file
Device: 259,9   Inode: 11947048    Links: 1
Modify: 2026-02-07 10:06:50.918485119 -0600
 Birth: 2026-02-07 10:06:50.918485119 -0600
> stat test.ouch.txt
  File: test.ouch.txt
  Size: 0               Blocks: 0          IO Block: 4096   regular empty file
Device: 259,9   Inode: 16384698    Links: 1
Modify: 2026-02-07 04:06:50.000000000 -0600
 Birth: 2026-02-07 10:07:18.766744263 -0600
```

### Expected Behavior

I would expect the same output that `unzip` gives:

```
> stat test.unzip.txt
  File: test.unzip.txt
  Size: 0               Blocks: 0          IO Block: 4096   regular empty file
Device: 259,9   Inode: 11962158    Links: 1
Modify: 2026-02-07 10:06:50.000000000 -0600
 Birth: 2026-02-07 10:07:31.198860767 -0600
```

### Additional Information

Here's the `unzip` version, in case it's relevant:

```
> unzip --version
UnZip 6.00 of 20 April 2009, by Debian. Original by Info-ZIP.
```

Thanks for `ouch`!!

[test.orig.txt](https://github.com/user-attachments/files/25153097/test.orig.txt)
[test.ouch.txt](https://github.com/user-attachments/files/25153095/test.ouch.txt)
[test.unzip.txt](https://github.com/user-attachments/files/25153094/test.unzip.txt)
[test.zip](https://github.com/user-attachments/files/25153096/test.zip)

Guflly:
Fixes #902.

ZIP stores modification times as timezone-naive local clock values. Ouch previously treated those values as UTC during extraction and wrote UTC clock values when creating archives, shifting file times by the local UTC offset.

Archive creation now converts filesystem timestamps to the applicable local offset before writing the ZIP timestamp. Extraction resolves the local offset for the stored date and converts that clock value back to the corresponding filesystem timestamp, including historical daylight-saving offsets.

Tests cover a fixed non-UTC offset and the host's actual local offset.

Validation:

- `cargo test --no-default-features --features unrar,use_zlib,use_zstd_thin` (103 passed)
- `rustfmt --edition 2024 --check --config newline_style=Native src/archive/zip.rs`
- `cargo clippy --bin ouch --no-default-features --features unrar,use_zlib,use_zstd_thin`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1122,6 +1122,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "num_threads"
+version = "0.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5c7398b9c8b70908f6371f47ed36737907c87c52af34c268fed0bf0ceb92ead9"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "once_cell"
 version = "1.21.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1767,7 +1776,9 @@ checksum = "18dfaaeddcb932337b5e7866ee7d0ce9b76d2fd092997146f187ec09b4558a50"
 dependencies = [
  "deranged",
  "js-sys",
+ "libc",
  "num-conv",
+ "num_threads",
  "powerfmt",
  "serde_core",
  "time-core",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ snap = "1.1.1"
 strum = { version = "0.28.0", features = ["derive"] }
 tar = "0.4.46"
 tempfile = "3.27.0"
-time = { version = "0.3.47", default-features = false }
+time = { version = "0.3.47", default-features = false, features = ["local-offset"] }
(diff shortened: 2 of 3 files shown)


valoq:
LGTM

~~Note, there should be a changelog entry for the fix~~

PR #1047, opened 2026-08-16
Document Debian installation via the pkg.haus APT archive

barnumbirr:
[pkg.haus](https://pkg.haus) is a signed APT archive carrying ouch for Debian stable, testing and unstable (amd64/arm64), built from source at upstream release tags. This adds a Debian section to the installation list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,14 @@ cargo install ouch
 
 (If you're on Ubuntu, you might need to install `clang` to build it from crates.io.)
 
+## On Debian via pkg.haus
+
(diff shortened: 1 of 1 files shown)


marcospb19:
Hey, thanks for packaging ouch.

While I'm skeptical of new projects like yours (not personal), I've reviewed your history in packaging and chose to trust your new registry  :) , I wish you success with pkg.haus, if you stick to it and manage to share/delegate some of the work with trustful people, I think it can grow and be great a thing similar to AUR or Homebrew (right now I see it's mostly you creating the first packages).

If it all works out I hope you can eventually get into the repology.org list so it'd appear in my README for free.

marcospb19 (approved):

PR #1049, opened 2026-08-17
fix: decompressing with --format overwrites the input file

### LINKED ISSUE #442: When decompressing with `--format`, don't overwrite input
When using the `--format` flag, `Ouch` will decompress the output while overwriting the input file.

```
 $ touch file                           
 $ ouch compress file file.zst.zst.zst
[INFO] Successfully compressed 'file.zst.zst.zst'.
 $ ouch decompress file.zst.zst.zst --format zst   
Do you want to overwrite './file.zst.zst.zst'? [Y/n] 
...
```

It would be good to decompress the file without erasing the previous data.

VXNCXNX:
## What's broken

`ouch decompress <file> --format <fmt>` derives the output name from the input's full file name, so the two are equal. With `--here` (or `--dir .`) the input is truncated to zero bytes and the run then fails, so the archive is gone.

Fixes #442

## Repro

```
$ echo "hello world data" > file
$ ouch compress file file.zst.zst.zst
$ ls -l file.zst.zst.zst
44 file.zst.zst.zst

$ ouch decompress file.zst.zst.zst --format zst --here -y
[ERROR] Failed to decompress file.zst.zst.zst
 - incomplete frame
$ ls -l file.zst.zst.zst
0 file.zst.zst.zst        <- input destroyed

after:
[INFO] Input file size: 44 B / Output file size: 35 B
44 file.zst.zst.zst       <- intact, md5 unchanged
35 file.zst.zst           <- output
```

## The fix

Strip as many trailing known extensions as `--format` names, so `file.zst.zst.zst --format zst` writes `file.zst.zst`. A name with no known extension falls back to `<name>-output` rather than colliding.

## Verification

`decompress_with_format_flag_does_not_overwrite_input` asserts the input bytes are unchanged and the derived output exists. Reverting the name derivation fails it: the input reads back as empty.

I also checked the paths this could have disturbed, comparing binaries built before and after: archive extraction with `--format`, stdin input, and plain `.gz` decompression without `--format` all behave identically.

Full suite passes. `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean. I built with `bzip3` off, since `libbzip3-sys` will not compile here (missing `stddef.h`), and nothing in the diff touches it.

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -187,7 +187,21 @@ pub fn run(args: CliArgs, question_policy: QuestionPolicy, file_visibility_polic
                     let file_name = path.file_name().ok_or_else(|| Error::Custom {
                         reason: FinalError::with_title(format!("{} does not have a file name", PathFmt(path))),
                     })?;
-                    files_output_paths.push(file_name.into());
+                    let output_file_name = match <[u8] as ByteSlice>::from_os_str(file_name) {
+                        Some(bytes) if !is_path_stdin(path) => {
+                            let stripped = extension::strip_known_extensions_from_name(bytes, format.len());
+                            if stripped == bytes {
+                                utils::append_ascii_suffix_to_os_str(file_name, "-output")
+                            } else {
+                                stripped
+                                    .to_os_str()
+                                    .expect("stripped bytes came from an OsStr")
+                                    .to_owned()
+                            }
+                        }
+                        _ => file_name.to_owned(),
+                    };
+                    files_output_paths.push(output_file_name.into());
                     files_extensions.push(format.clone());
                 }
             } else {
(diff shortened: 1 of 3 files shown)


VXNCXNX:
Rebased onto current `main`. The only conflict was `CHANGELOG.md`, and `main` has since marked it "This changelog shouldn't be used anymore. This file will be removed soon.", so I dropped my entry rather than re-adding it. `CHANGELOG.md` here is now identical to `main`; the diff is code and tests only.

On the new base: `cargo test --bins` and `cargo test --test integration` pass, clippy has no errors.

marcospb19:
Hey, thanks for this PR, I apologize for the delay, I'll be reviewing soon.

PR #1050, opened 2026-08-17
fix: bz3 is missing from SUPPORTED_EXTENSIONS

VXNCXNX:
## What's broken

`bz3` is missing from `SUPPORTED_EXTENSIONS`, although `slice_to_extension` parses it and `PRETTY_SUPPORTED_EXTENSIONS` advertises it. The "insert an archive format" hint walks that list, so it does not see the `bz3` boundary and suggests a path ouch then refuses.

## Repro

```
$ ouch compress folder out.bz3.gz
before:  hint: To:   "out.bz3.tar.gz"
after:   hint: To:   "out.tar.bz3.gz"
```

Following the old hint verbatim:

```
$ ouch compress folder out.bz3.tar.gz
[ERROR] File extensions are invalid for operation
 - The archive extension '.tar' can only be placed at the start of the extension list
```

So the hint pointed at a path ouch rejects on its own terms.

## The fix

One entry added to `SUPPORTED_EXTENSIONS`.

## Verification

Two tests. One pins the suggestion for `linux.bz3` and `linux.bz3.gz`. The other walks every extension in `PRETTY_SUPPORTED_EXTENSIONS`, skips the archive-first ones, and asserts the rest are in `SUPPORTED_EXTENSIONS`, so a future addition cannot go missing the same way. Removing `bz3` again fails both, the second with the name of the offending extension.

Output above is from binaries built before and after, not from the unit tests.

Full suite passes. `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean. I built with `bzip3` off, since `libbzip3-sys` does not compile here (missing `stddef.h`); that only affects whether the resulting archive can be written, not the extension bookkeeping this changes.

`tbz2` and `tbz3` are parsable but absent from `SUPPORTED_ALIASES`. They are archive-first, so the suggestion path never consults them, and I left them alone to keep this to one concern.

### DIFF
--- a/src/extension.rs
+++ b/src/extension.rs
@@ -16,6 +16,7 @@ pub const SUPPORTED_EXTENSIONS: &[&str] = &[
     "zip",
     "bz",
     "bz2",
+    "bz3",
     "gz",
     "lz4",
     "xz",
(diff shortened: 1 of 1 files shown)


VXNCXNX:
Rebased onto current `main`. The only conflict was `CHANGELOG.md`, and `main` has since marked it "This changelog shouldn't be used anymore. This file will be removed soon.", so I dropped my entry rather than re-adding it. `CHANGELOG.md` here is now identical to `main`; the diff is code and tests only.

On the new base: `cargo test --bins` and `cargo test --test integration` pass, clippy has no errors.

src/extension.rs
```diff
@@ -406,6 +407,40 @@ mod tests {
...
+    /// `ouch` itself rejects. This is what let `bz3` go missing.
+    /// Archive-first aliases such as `tgz` and `tbz3` live in `SUPPORTED_ALIASES` instead.
+    #[test]
+    fn every_standalone_parsable_extension_is_supported() {
+        for ext in PRETTY_SUPPORTED_EXTENSIONS.split(", ") {
+            let parsed =
+                slice_to_extension(ext.as_bytes()).unwrap_or_else(|| panic!("'{ext}' is advertised but not parsable"));
+            if parsed.compression_formats == [CompressionFormat::Tar] {
+                continue;
+            }
```
marcospb19, line 436: Hi, why are we skipping tar here?

VXNCXNX: Nothing, and you are right to ask. `tar` is itself in `SUPPORTED_EXTENSIONS`, so that guard was skipping a case that already passed the assertion. It was dead weight and it narrowed the test for no reason.

Removed in 23ff69e. The loop now covers `tar` as well, so dropping it from the list fails too:

```
'tar' is advertised and parsable but missing from SUPPORTED_EXTENSIONS
```

Before the change, that mutation passed. `bz3` is still caught, which was the original point. Sorry for the slow reply on this one.

Issue #1005, opened 2026-06-04
listing: add flag to show file sizes


marcospb19:
Add a flag like `ouch list archive.tar.gz --show-size` that shows the file size for each listed entry.

This requires that we read the contents of each entry instead of just their name, so it's more computationally heavy, adding this flag shouldn't make the default behavior slower.

_Originally posted by @ofek in [#1000](https://github.com/ouch-org/ouch/issues/1000#issue-4529358992)_

marcospb19:
I think it's OK if this is not the default, as it pollutes the output and we'll be adding the `ouch info` command suggested in #1000, which should already tell the archive size.

PR #1052, opened 2026-08-18
Add --show-size flag to `ouch list` (#1005)

ChrisJr404:
This adds the `--show-size` flag to `ouch list` that was asked for in #1005, so you can see how big each entry is without extracting anything.

The sizes come straight from the archive metadata (tar/zip/7z/rar all report the uncompressed size per entry), so the plain `ouch list` output is byte-for-byte the same as before and you only get the extra column when you pass the flag. Files show a right-aligned size using the same `BytesFmt` the rest of ouch uses; directories get a blank, aligned column so things line up. It works in `--tree` mode too, with the size column sitting to the left of the branches.

```
$ ouch list proj.tar.gz --show-size
Archive: "proj.tar.gz"
            proj/
 12.00   B  proj/README.md
            proj/src/
  2.00   B  proj/src/tiny.txt
  3.00 kiB  proj/src/big.bin
```

Added an integration test covering both the flag and the unchanged default output.

### DIFF
--- a/src/list.rs
+++ b/src/list.rs
@@ -10,7 +10,7 @@ use self::tree::Tree;
 use crate::{
     Result,
     accessible::is_running_in_accessible_mode,
-    utils::{NoQuotePathFmt, PathFmt},
+    utils::{BytesFmt, NoQuotePathFmt, PathFmt},
 };
 
 /// Options controlling how archive contents should be listed
@@ -21,6 +21,9 @@ pub struct ListOptions {
 
     /// Whether to suppress extra output like symlink targets (for scripting)
     pub quiet: bool,
+
+    /// Whether to show the (uncompressed) size of each file
+    pub show_size: bool,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -39,6 +42,9 @@ pub struct FileInArchive {
 
     /// The type of file
     pub file_type: ListFileType,
+
+    /// The uncompressed size of the entry, when the format reports it
+    pub size: Option<u64>,
 }
 
 /// Actually print the files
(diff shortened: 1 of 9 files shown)


src/list.rs
```diff
@@ -56,16 +62,38 @@ pub fn list_files(
...
+
+    match (file_type, size) {
+        (ListFileType::File | ListFileType::Symlink { .. } | ListFileType::Hardlink { .. }, Some(size)) => {
+            let _ = write!(out, "{}  ", BytesFmt(size));
+        }
+        _ => {
+            // `BytesFmt` renders as 10 columns, keep the padding in sync with it.
+            let _ = write!(out, "{:>10}  ", "");
+        }
+    }
```
marcospb19: Small nitpick from me:


```suggestion
        if let Some(size) = size
            && !matches!(file_type, ListFileType::Directory { .. })
        {
            let _ = write!(out, "{}  ", BytesFmt(size));
        } else {
            // `BytesFmt` renders as 10 columns, keep the padding in sync with it.
            let _ = write!(out, "{:>10}  ", "");
        }
```

marcospb19 (changes requested):
Great code and great PR, thanks.

Cause I merged older PRs before, now you have conflicts, just 1 nitpick the rest is perfect.

ChrisJr404:
Thanks! Applied your nitpick (the `if let Some(size) = size && !matches!(file_type, ListFileType::Directory)` form; dropped the `{ .. }` since `Directory` is a unit variant, which clippy also prefers).

I also merged `main` to resolve the conflicts with the recently-merged `--depth` work. `--show-size` and `--depth` now compose: the size column is rendered inside the depth-limited traversal in both flat and tree views. Build, fmt, clippy, and both the `list_show_size` and `list_depth` integration tests pass locally.

marcospb19 (approved):
Great, thank you so much.

PR #1054, opened 2026-08-21
add audit ci workflow

valoq:
This workflow runs cargo audit weekly and on dependency changes and opens a PR for any findings.

This may create more bot PRs though.

The alternative is to just run the audit without PRs and run `cargo update` manually before releases, which is more error prone

### DIFF
--- a/.github/workflows/security-audit.yml
+++ b/.github/workflows/security-audit.yml
@@ -1,0 +1,49 @@
+name: Security audit
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "**/Cargo.toml"
+      - "**/Cargo.lock"
+  pull_request:
+    paths:
+      - "**/Cargo.toml"
+      - "**/Cargo.lock"
+  schedule:
+    - cron: "0 6 * * 1"
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  cargo-audit:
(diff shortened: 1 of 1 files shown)


marcospb19:
if the bot opens a PR, that catches our attention, that's important at least

so would it be reasonable for me to close it, bump that dependency myself and make a PR for it?

marcospb19:
maybe it would be better to run `cargo update dep` instead of the more general `cargo update` :thinking: 

if that's the case I think I'm fine with the bot PRs : ) 

wdyt

valoq:
What most rust projects do is just to run the audit command in ci and have the ci report failures if something comes up.
We could do that too but it would add more work.

Im not sure whats the best option here

marcospb19:
If that's harder, I'd keep this approach.

About my previous suggestion, maybe we could do:

```
cargo audit --json | jq -r '.vulnerabilities.list[].package.name' | tr '\n' ' '
```

To get the list of packages that were found and then we pass that down from one step to the other, and invoke `cargo update $(pkgs)` instead of the wider `cargo update`.

EDIT: using the $GITHUB_OUTPUT feature.

valoq:
Its not harder, its more manual work, but I suspect it will rarely fire.
I will open another PR to just add the ci check for now and we run `cargo update` or `cargo update dep` manually when it fires.
dependabot should already catch most.

PR #1056, opened 2026-08-25
Add --depth option to the list command

### LINKED ISSUE #638: Request: Add Depth Option to `--tree` for Listing Archive Contents
Hello `ouch` team,

First of all, thank you for developing such a practical and user-friendly tool. `ouch` has been incredibly useful, especially the `list` command's `--tree` option, which displays the contents of archives in a tree structure. This feature is immensely helpful for quickly understanding the structure of an archive.

However, I've noticed that the `--tree` option seems to lack the ability to limit the display depth, similar to the `-L` option in the `tree` command. This capability would be particularly useful when dealing with archives that contain a large number of nested directories.

**Feature Request:**
I would like to request the addition of an option to the `list --tree` command to specify the maximum depth of display, such as `list --tree 2`, which would only show up to the second level of directories and files.

I believe this feature would make `ouch` even more powerful and flexible. Thank you again for your hard work, and I look forward to seeing `ouch` continue to evolve!

Best regards,

ChrisJr404:
Closes #638.

Adds a `--depth N` flag to `ouch list` that only shows entries down to N levels, like the issue asked for. Works the same for the flat listing and for `--tree`, top-level entries count as depth 1.

```
$ ouch list src.tar.gz --tree --depth 2
Archive: "src.tar.gz"
└── src/
   ├── top.txt
   └── a/
```

Went with a single `--depth` since that seemed to be the preference in the issue, no `--min-depth`/`--max-depth`. Default behavior is unchanged when the flag isn't passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,15 +16,11 @@ Categories Used:
 
 **Bullet points in chronological order by PR**
 
-## [Unreleased](https://github.com/ouch-org/ouch/compare/0.8.1...HEAD)
+# For any versions higher than 0.8.1
 
-### New Features
+This changelog shouldn't be used anymore.
 
-- Add `--depth` to `list` to only show entries up to a given number of levels (https://github.com/ouch-org/ouch/pull/1056).
-
-### Tweaks
-
-- Releases: sign assets with cosign instead of GitHub artifact attestations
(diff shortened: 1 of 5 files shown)


marcospb19:
Thanks!!

PR #1057, opened 2026-08-25
fix(completions): clarify compress positional arguments

### LINKED ISSUE #979: Shell completion descriptions may mislabel input files and output archive
### Version

0.7.1

### Description

Shell completions for `ouch compress` can show misleading completion descriptions.

### Current Behavior

When I'm typing `ouch compress ` (a space in the end), the hint shows:

```sh
~/Projects/test
> ls
file.txt

~/Projects/test
> ouch compress 
output -- The resulting file. Its extensions can be used to specify the compression formats // completion desc
```

Then, I continue typing `ouch compress file.txt ` (a space in the end):

```sh
~/Projects/test
> ouch compress file.txt
files -- Files to be compressed // completion desc
```

This is confusing because users expect `ouch compress <FILES>... <OUTPUT>`, but completion description can make it look as `ouch compress <OUTPUT> <FILES>...`.

### Expected Behavior

Completion descriptions should reflect the documented argument order: `ouch compress <FILES>... <OUTPUT>`. They should not suggest that the output archive comes before the input files.

### Likely Cause

CAUTION: The content in this section is suggested by `openai/gpt-5.5`.

```rust
// src/cli/args.rs
Subcommand::Compress {
    /// Files to be compressed
    #[arg(required = true, value_hint = ValueHint::FilePath)]
    files: Vec<PathBuf>,

    /// The resulting file. Its extensions can be used to specify the compression formats
    #[arg(required = true, value_hint = ValueHint::FilePath)]
    output: PathBuf,
}
```

A variadic positional argument followed by another required positional is parseable once the full command is known, but it is inherently hard for shell completion because the command line is incomplete while completion is being requested.

`build.rs` appears to only expose this via generated completions:

```rust
for shell in Shell::value_variants() {
    generate_to(*shell, cmd, "ouch", out).unwrap();
}
```

### Possible Fixes

CAUTION: The content in this section is suggested by `openai/gpt-5.5`.

1. Prefer an explicit output option for future CLI shape, e.g.

   ```sh
   ouch compress file1 file2 --output archive.zip
   ```

   This removes the ambiguity for completions.

2. If preserving the current syntax is required, add custom completion behavior for `compress` so completions do not present misleading `files`/`output` labels.

3. At minimum, adjust generated/manual completion text to avoid suggesting that `output` comes before input files.

SeojunKim-pumisj:
Generated shell completions describe `<FILES>...` and `<OUTPUT>` independently, which can make their positional order look reversed while the command line is incomplete.

Use a shared description for both generated completion entries. `--help` and man-page descriptions remain unchanged.

Fixes #979

Tested with:
- `cargo test`
- `cargo clippy -- -D warnings`
- generated Zsh and Nushell completion artifacts

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -18,6 +18,8 @@ use clap_complete::{Shell, generate_to};
 use clap_complete_nushell::Nushell;
 
 include!("src/cli/args.rs");
+#[path = "src/cli/completion.rs"]
+mod completion;
 
 fn main() {
     println!("cargo:rerun-if-env-changed=OUCH_ARTIFACTS_FOLDER");
@@ -29,6 +31,7 @@ fn main() {
 
         clap_mangen::generate_to(cmd.clone(), out).unwrap();
 
+        let cmd = &mut completion::with_combined_compress_positionals(CliArgs::command());
         for shell in Shell::value_variants() {
             generate_to(*shell, cmd, "ouch", out).unwrap();
         }
--- a/src/cli/completion.rs
+++ b/src/cli/completion.rs
@@ -1,0 +1,64 @@
+use clap::Command;
+
+const COMPRESSION_POSITIONAL_HELP: &str = "Input files; or output archive with compression format when placed last";
+
+pub fn with_combined_compress_positionals(mut command: Command) -> Command {
+    let compress = command
+        .find_subcommand_mut("compress")
+        .expect("compress subcommand should exist");
+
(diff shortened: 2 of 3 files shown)


marcospb19:
> generated Zsh and Nushell completion artifacts

Thanks for the PR and for testing it, appreciate it!

PR #1063, opened 2026-09-01
fix: support Zstandard (method 93) entries in zip archives

### LINKED ISSUE #1062: ouch list and decompress fail on zip archives using Zstandard compression (method 93)
### Version

0.8.2

### Description

ouch list and ouch decompress fail on zip archives whose entries are compressed with Zstandard (compression method 93) "zstd"

### Current Behavior

Create a zstd zip and list it or using any other tool to compress a zip using zstd compression:
- printf 'hello\n' > a.txt
- bsdtar --options zip:compression=zstd -cf test.zip a.txt

ouch list test.zip
Output:
Archive: "test.zip"

[ERROR] Unexpected error in zip archive
 - compression method not supported: 93

ouch decompress test.zip fails the same way.

### Expected Behavior

ouch list and ouch decompress should handle method-93 (Zstandard) zip entries like any other zip.

### Additional Information

OS: Arch Linux (CachyOS). 
Root cause: ouch uses the Rust zip crate (8.6.0) with the zstd feature disabled, so the crate maps method 93 to Unsupported. 
Enabling that feature in Cargo.toml fixes both list and decompress

7487:
Fixes #1062.

`ouch list` and `ouch decompress` failed on zip archives whose entries use Zstandard compression (method 93) with `compression method not supported: 93`, because the `zip` crate was built without its `zstd` feature.

This enables the feature — it resolves to the same `zstd` 0.13.3 already in the dependency tree, so the lockfile change is a single line — and adds a regression test that lists and decompresses a zip with a Zstd-compressed entry. Without the fix the test fails to compile, since `CompressionMethod::Zstd` is gated behind that feature.

Tested with `cargo test` (all suites pass) and `cargo fmt --check`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2095,6 +2095,7 @@ dependencies = [
  "time",
  "typed-path",
  "zeroize",
+ "zstd",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ unrar = { package = "unrar-ng", version = "0.7.6", optional = true }
 zip = { version = "8.6.0", default-features = false, features = [
     "time",
     "aes-crypto",
+    "zstd",
 ] }
 zstd = { version = "0.13.3", default-features = false, features = ["zstdmt"] }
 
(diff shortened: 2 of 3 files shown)


marcospb19 (approved):
Checked with `ldd` and `du -h` locally, no changes.

Thanks!

PR #1064, opened 2026-09-02
Prevent overwriting during compression

valoq:
Fixes the overwrite issue during compression as well


marcospb19 (approved):
let's gooooooo 🔥 that's so great

PR #1080, opened 2026-09-16
fix: crash on --format

valoq:
When decompress with --format using the wrong order, ouch crashes

poc:
`ouch decompress backup.tar.gz --format gz.tar`


marcospb19 (approved):

marcospb19:
ty

